PR #18, opened 2023-03-11
Create api

nh916:
# Description

setting up the api client

* Started setting up the API object and added some placeholder tests. 
* Cached API object as global api
* Everything has type hints
* All methods and tests have docstrings.
* Added fetch schema and validate schema into functions.
* Added fetch controlled vocabulary and validate controlled vocabulary placeholder functions
    * Created Exception for Invalid vocabulary and placeholder link

## Changes
* added API object
* the `_get_db_schema()` method works
* the rest of the methods are just empty with instructions of what needs to go there
* added tests for the API object
* Added a small requirements.txt file
* Added mypy as a dependency to check our type hints for our code and catch any errors that might come up

## Known Issues
* docstrings will need to eventually be improved to be user friendly later when we get to the documentation stage
* we talked about caching the API, I have not done that currently, I am still debating if we can only cache the host.
    * We might be putting the cart before the horse here in that we want to ensure that the SDK can work with a private instance when I am not sure yet if the API can work as a private instance yet, however, I think our idea should be okay
* all methods are empty
* all tests are empty
* `test_api_search_material_by_url` is potentially testing two things, which could not be ideal

## Unsure about
I added a list of `_VALID_SEARCH_MODES` in the `_valid_search_modes.py` file, and I am unsure if that is the best way that it can be done, if it is scalable, or if needed. I think it is a pretty clean way of going about it in my opinion, but not 100% sure.

## Note
After the [Make Cript installable via pip PR](https://github.com/C-Accel-CRIPT/Python-SDK/pull/16) is submitted, there is a chance that this will have a merge conflict because it is not in the `cript/` directory, but we can fix that pretty easily with a `git merge develop`


src/cript/api/api.py
```diff
@@ -0,0 +1,164 @@
...
+            warnings.warn("HTTP is an unsafe protocol please consider using HTTPS")
+
+        # check that api can connect to CRIPT with host and token
+        try:
+            # TODO send an http request to check connection with host and token
+            pass
+        except Exception:
+            raise CRIPTConnectionError
+
+    def save(self, node: PrimaryBaseNode) -> None:
```
CVilla17, line 64: Even though it is fairly self explanatory, I would add a docstring of the exact specification for each function

CVilla17 (commented):
I think this all looks great so far, I had one comment about making sure specs were on all functions but it seems that for the most part that is accounted for

InnocentBug (approved):

src/cript/api/api.py
nh916: that is completely fair, and I fully agree. I added a brief docstrings for this but it will likely need more explanation for a graph that is both has nodes to both save and update, but we'll add that when we get to it

src/cript/api/vocabulary.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # convert to dict for easier use
+        response = json.loads(response)
+
+        _ENTIRE_CONTROLLED_VOCABULARY = response
+
+    return _ENTIRE_CONTROLLED_VOCABULARY
+
+
+def is_vocab_valid(vocab_category: str, vocab_value: str) -> Union[bool, InvalidVocabulary]:
```
bearmit, line 43: This is not supposed to wrap a schema validation function? You have the schema for the vocab, you can check it using your schema library.

src/cript/api/api.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def _get_global_cached_api():
+    """
+    Read-Only access to the globally cached API object.
+    Raises an exception if no global API object is cached yet.
+    """
+    if _global_cached_api is None:
+        raise CRIPTAPIAccessError
+    return copy.copy(_global_cached_api)
```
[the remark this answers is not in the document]
InnocentBug: Python doesn't really have the concept of constant objects.
Not like C++ for example.
So, I can't really make it constant.

The best approximation we have is to return a copy, since there is never write access to the globally cached instance.

src/cript/api/schema_validation.py
```diff
@@ -0,0 +1,62 @@
...
+    1. Checks if the class variable is already set and if it is then it just returns that.
+    2. If db schema is not already saved, then it makes a request to get it from CRIPT
+    3. after successfully getting it from CRIPT, it sets the class variable
+
+    Returns
+    -------
+    json
+        The database schema in JSON format.
+    """
+    host: str = "https://development.api.criptapp.org"
```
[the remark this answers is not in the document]
InnocentBug: Yes that was from pre caching the API @nh916 can you change it?

src/cript/api/vocabulary.py
```diff
@@ -0,0 +1,84 @@
...
+        return True
+
+    controlled_vocabulary: dict = _get_controlled_vocabulary()
+
+    try:
+        if vocab_value in controlled_vocabulary[vocab_category]:
+            return True
+        else:
+            # if the vocabulary does not exist in a given category
+            raise InvalidVocabulary
```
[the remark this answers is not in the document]
InnocentBug: This is a dummy for now.

Once we actually implement this properly the exception will have more details.

src/cript/api/vocabulary.py
InnocentBug: Yes, that is what is going to happen under the hood there.

brili (approved):

CVilla17 (commented):
Berenger has some good points that have been discussed, so pending resolution of those points I would approve because this is generally a good framework

src/cript/api/api.py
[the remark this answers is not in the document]
nh916: I looked into making the function static a while ago and from what I remembered it didn't make sense to have the host be dynamic, but the `_get_host()` be static. However, if I am wrong this could be a good way to go about it, by just making a single method static that always gets the host and the `validate_schema` and `validate_vocab` can just get the host and make their own calls to the API to get the needed resources and return either a True or return an Exception.

src/cript/api/api.py
```diff
@@ -0,0 +1,234 @@
...
+        None
+        """
+        # TODO create a giant JSON from the primary node given and send that to
+        #   the backend with a POST request
+        #   the user will just hit save and the program needs to figure out
+        #   the saving of new or updating
+        #   save is POST request and update would be PATCH request
+        pass
+
+    def delete(self, node: PrimaryBaseNode, no_input: bool = False) -> None:
```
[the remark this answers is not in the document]
nh916: I was originally thinking something like `python manage.py collectstatic --no-input`, but I think `ask_confirmation = False` is probably better

I made the change

src/cript/api/schema_validation.py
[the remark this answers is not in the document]
nh916: yup, done!

src/cript/api/schema_validation.py
```diff
@@ -0,0 +1,62 @@
...
+    Returns
+    -------
+    bool
+        whether the node JSON is valid or not
+    """
+
+    db_schema = _get_db_schema()
+
+    # TODO currently validate says every syntactically valid JSON is valid
+    # TODO do we want invalid schema to raise an exception?
```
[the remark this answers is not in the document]
nh916: I think it's too early and for now I can just either return True or raise an error, but after we start messing around with it and if we see we have the need for that or that would just feel better then we can add it.

src/cript/api/vocabulary.py
```diff
@@ -0,0 +1,84 @@
...
+    -------
+    dict
+        controlled vocabulary
+    """
+
+    global _ENTIRE_CONTROLLED_VOCABULARY
+
+    # if cache is empty then make request and set cache
+    if len(_ENTIRE_CONTROLLED_VOCABULARY) == 0:
+        host: str = "https://development.api.criptapp.org"
```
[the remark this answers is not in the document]
nh916: done!

src/cript/api/vocabulary.py
nh916: I was thinking we we're going to get the controlled vocabulary from the API instead of the db schema.

with something like this

```
[
algorithm_key, algorithm_type, building_block, citation_type, computation_type, computational_forcefield_key, ...
]

# for loop to make a request to the api for each end point and put the result inside the _ENTIRE_CONTROLLED_VOCABULARY

```

and when we want to check a vocabulary value like data type, we would call the function with the category and word

```
is_vocab_valid("data_type", "afm_amp")
```

src/cript/api/api.py
```diff
@@ -0,0 +1,234 @@
...
+from typing import List, Literal, Union
+
+from cript.api._valid_search_modes import _VALID_SEARCH_MODES
+from cript.api.exceptions import CRIPTAPIAccessError, CRIPTConnectionError
+from cript.nodes.primary_nodes.primary_base_node import PrimaryBaseNode
+from cript.nodes.primary_nodes.project import Project
+from cript.nodes.supporting_nodes.group import Group
+from cript.nodes.supporting_nodes.user import User
+
+_global_cached_api = None
```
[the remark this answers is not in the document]
InnocentBug: It is not a singleton, since you can change the API access during the execution of your python program.
Instance might work, but I like cache better, since it just caches an otherwise existing object.

InnocentBug (dismissed):
This is just a start and we don't really test things yet.
But let's get it in, that we have something to work with.

src/cript/api/schema_validation.py
```diff
@@ -21,7 +23,7 @@ def _get_db_schema() -> json:
     json
         The database schema in JSON format.
     """
-    host: str = "https://development.api.criptapp.org"
+    host: str = _global_cached_api.host
```
bearmit, line 26: _global_cached_api can't be null?

InnocentBug: no it can.
@nh916 please use `_get_global_cache_api()` instead.
That catches that problem.

good catch @bearmit

InnocentBug (approved):
I don't love the design of how the vocab and schema is cached globally.
I think capsulating that into the API is better and prevents problems down the road.

For now this is OK, though.

src/cript/api/schema_validation.py
```diff
@@ -0,0 +1,64 @@
+import json
+
+import requests
+from jsonschema import validate
+
+from cript.api.api import _get_global_cached_api
+
+_DB_SCHEMA: json = None
```
InnocentBug, line 8: This should be `_DB_SCHEMA: dict` right now it is a module type json.

Actual json data is dicts in python.

InnocentBug: done

src/cript/api/schema_validation.py
nh916: Thank you so much @InnocentBug for updating this and making the push for me! 🙏

InnocentBug (approved):

bearmit (approved):

src/cript/api/api.py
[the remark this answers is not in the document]
dylanwal: I not of fan of using copy here, I would just return the object itself. If a users messing with a private object (object with leading underscore) they do so knowing the will possibly break the code (this is Python general practice).  
If you do want to make it constant; it is typically done by making it a frozen class.

src/cript/api/exceptions.py
```diff
@@ -0,0 +1,61 @@
+class CRIPTConnectionError(Exception):
```
dylanwal, line 1: you should make a base 'CRIPTException' class and have these error class inherit from that. This makes it easier for future work to catch all exceptions create by CRIPT SDK

src/cript/api/schema_validation.py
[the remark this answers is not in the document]
dylanwal: I would say yes; but can this validation catch all the errors an return it in one go, or does it stop after the first error? I think would be best to collect them all and raise an error provide a full list of issues for the user to fix.

src/cript/api/api.py
```diff
@@ -0,0 +1,234 @@
...
+                raise RuntimeError(
+                    "API initilized with `token=None` but environment variable `CRIPT_TOKEN` not found.\n"
+                    "Set the environment variable (preferred) or specify the token explictly at the creation of API."
+                )
+        # strip ending slash to make host always uniform
+        host = host.rstrip("/")
+
+        # if host is using unsafe "http://" then give a warning
+        if host.startswith("http://"):
+            warnings.warn("HTTP is an unsafe protocol please consider using HTTPS.")
```
dylanwal, line 90: I would recommend using pythons logging module; its a lot more flexible and better as things get more complex. Its also nice for people who are going to build upon this package for getting log reports for debugging.

dylanwal (approved):
I provided some recommendations but mostly looks good

src/cript/api/api.py
InnocentBug: There is actually a page that lists, how to log errors in python and which module to use.
https://docs.python.org/3/howto/logging.html

To me it seems like that `warnings.warn` is the right decision here.
The "client application" can do something about this: The user script can specify https instead of http.
So warnings is appropriate.

PR #24, opened 2023-03-16
Implement a local search for nodes

InnocentBug:
# Description
This implementation allows the local search of nodes inside an existing tree of nodes.
It does not fetch new nodes, just helps to find existing nodes.

## Changes
Implementation of one public function in BaseNode. and a helper function.

## Known Issues
If we have cycles in our node graphs, this can potentially hang.

The shown example with deepcopy doesn't work right now.
We have to overwrite the deepcopy function for nodes, that removes the url and uid and other objects from the the node, otherwise the backend would not treat them as separate nodes.

The search is written as a recursion with the inherent drawbacks, could be optimized. 

## Notes

This can be very powerful in the context of duplicating a sub-graph and changing some values.
For example if you redid an experiment, but for example the pressure condition changes, you can do this.
```python
import copy
experimentA = get_original_exp()
experimentB = copy.deepcopy(experimentA)

#Now let's change the pressure everywhere.
pressure_cond = experimentA.find_children({"node": "Condition", "name": "pressure"})
for pc in pressure_cond:
    pc.value = 2.0
# Now change the resulting properties of the material etc too.
experimentA.find_children({"node": "Property"})

### CODE REVIEW

File: tests/test_nodes_no_host.py [hunk 1]
```diff
@@ -52,3 +52,34 @@ def test_algorithm():
     assert a.type == "integration"
 
     #Add citation test, once we have citation implemted
+
+def test_search():
```
nh916, line 56: I recommend docstrings for this test to explains its purpose and what it tries to test for

src/cript/nodes/core.py
```diff
@@ -72,3 +73,73 @@ def json(self):
...
+                            number_values_found += 1
+                            break
+            return number_values_found == len(value)
+
+
+        found_children = []
+
+        # Handle self
+        found_attr = 0
+        for key, value in search_attr.items():
```
nh916, line 127: can you add some comments

src/cript/nodes/core.py
```diff
@@ -72,3 +73,73 @@ def json(self):
...
+            attr_key = asdict(node._json_attrs).get(key)
+
+            # To save code paths, I convert non-lists into lists with one element.
+            if not isinstance(attr_key, list):
+                attr_key = [attr_key]
+            if not isinstance(value, list):
+                value = [value]
+
+            number_values_found = 0
+            for v in value:
```
nh916, line 110: can you add some comments

src/cript/nodes/core.py
```diff
@@ -72,3 +73,73 @@ def json(self):
...
 
         return json.dumps(self, cls=NodeEncoder)
+
+    def find_children(self, search_attr:dict, recursion_depth:int=-1) -> List:
+        """
+        Finds all the children in a given tree of nodes (specified by its root),
+        that match the criteria of search_attr.
+        If a node is present multiple times in the graph, it is only once in the search results.
+
+        recursion_dept: Max depth of recursion into the tree. Helpful if circles are expected. -1 specifies no limit
```
nh916, line 83: I recommend removing the word `recursion` from docstrings because it could give a future developer an idea that this is using recursion when it is using iteration.

nh916 (approved):

nh916:
I strongly recommend implementing the changes I asked for, but I don't want to block development so I approved it, but if you do implemented the changes I asked for feel free to message me and I'll immediately review and to speed us up!

src/cript/nodes/core.py
```diff
@@ -72,3 +73,73 @@ def json(self):
...
+        found_attr = 0
+        for key, value in search_attr.items():
+            if is_attr_present(self, key, value):
+                found_attr += 1
+        if found_attr == len(search_attr):
+            found_children += [self]
+
+        # Handle recursion
+        if recursion_depth != 0:
+            for field in self._json_attrs.__dataclass_fields__:
```
nh916, line 135: I think this works perfectly fine for now, and it is exactly what we need to keep the code simple and good for the first version. In the future we probably would want to implement some other kind of search algorithms to make this search faster and more efficient. Like if we could make it into binary search then we could reduce the time complexity all the way down to `O(Logn)` which is just beautiful! But for first version I do not want to make this complicated at all.

nh916 (approved):

tests/test_nodes_no_host.py
InnocentBug: done

src/cript/nodes/core.py
InnocentBug: done

src/cript/nodes/core.py
InnocentBug: done

src/cript/nodes/core.py
InnocentBug: done, replaced it by search depth

src/cript/nodes/core.py
InnocentBug: Yeah, it is possible.
We would have to make the nodes sortable (so implementing a > condition). It is certainly possible by using a hash value from the json and make those sortable.
However, the search condition must understand this order then too. I don't think that would work with hashes though. No sure. Maybe someone smarter then knows something here.
The trouble is that, the specified conditions can be partial, which also rules out hash table search.

Maybe it is possible to implement some variation of bucket sort, where we sort the nodes in based on present attributes.
That would be constant in the number of nodes, better then log(N) and linear in the number of search criterions.
That is probably the best way to do it. But maybe something for the back end, rather then SDK.
It would also require a separate search structure that is build as nodes are loaded into the SDK, which is way too much overhead for the SDK.

brili (approved):

src/cript/nodes/core.py
nh916: yeah I agree, lets come back to it in future versions

nh916 (approved):

nh916 (approved):

nh916 (dismissed):

brili (approved):

nh916 (approved):

PR #25, opened 2023-03-16
Implement deepcopy for nodes

InnocentBug:
# Description

A deep copy of a node means to remove the url and uid from the node if present.
Otherwise the back end would see them as the same nodes.

## Known Issues
No testing yet, this needs at least one primary node implemented with a url to be tested.

## Notes


brili (approved):

tests/test_nodes_no_host.py
```diff
@@ -62,3 +62,47 @@ def test_algorithm():
     assert a.type == "integration"
 
     # Add citation test, once we have citation implemted
+
+
+def test_local_search():
```
CVilla17, line 67: I would split up the different tests that comprise this test for clarity

src/cript/nodes/core.py
```diff
@@ -73,3 +90,90 @@ def json(self):
...
+                                           Since parameter is a list an alternative notation is
+                                           ``{"node": "Algorithm", "parameter": [{"name" : "update_frequency"}]}`
+                                           and Algorithms are not excluded they have more paramters.
+                    search_attr = `{"node": "Algorithm", "parameter": [{"name" : "update_frequency"},
+                                           {"name" : "cutoff_distance"}]}`
+                                           finds all algoritms that have a parameter "update_frequency" and "cutoff_distance".
+
+        """
+
+        def is_attr_present(node: BaseNode, key, value):
```
CVilla17, line 119: If this ends up being applicable code to other functions in this file I would take this out of the scope of just this function, but it looks fine for now. Additionally, I believe it doesn't necessarily need to be in the function in the first place other than to try to hide its existence from the public, because it doesn't use anything native to the function in the way it is written

tests/test_nodes_no_host.py
InnocentBug: You mean having a separate test file for the search?
Yeah, that is an idea.

nh916 (approved):

InnocentBug:
Implementing this in #72

PR #36, opened 2023-03-21
fixed typo for `Dependencies: JSONSchema`

nh916:
# Description
Fixed typo for badges: 
* `Dependencies: JSONSchema`
* `Dependencies: Requests`
* `Dependencies: mypy`


InnocentBug (approved):

bearmit (approved):
There is no file that declares the dependencies? What's the purpose of those badges?

nh916:
> There is no file that declares the dependencies? What's the purpose of those badges?

@bearmit the file that shows the dependencies is 
* requirements.txt
* requirements_dev.txt
* requirements_docs.txt

I put the badges as attribution to the software that we are using. I figured that since we are open source attribution is important for us, and we can give some attributions of all the software that we are using and if anyone wants to contribute they have a glimpse immediately into what technologies we are using and how they can contribute

PR #41, opened 2023-03-23
Implementation of all remaining subobjects

InnocentBug:
# Description
This implements all remaining subobjects.

## Changes
Implements ComputationForcefield, Equipment, Condition, Property.

## Checklist

### DIFF
--- a/src/cript/nodes/subobjects/computation_forcefield.py
+++ b/src/cript/nodes/subobjects/computation_forcefield.py
@@ -1,0 +1,127 @@
+from dataclasses import dataclass, field, replace
+from typing import List, Union
+
+from cript.nodes.core import BaseNode
+from cript.nodes.primary_nodes.data import Data
+from cript.nodes.subobjects.citation import Citation
+
+
+class ComputationForcefield(BaseNode):
+    """
+    ComputationForcefield
+    """
+
+    @dataclass(frozen=True)
+    class JsonAttributes(BaseNode.JsonAttributes):
+        node: str = "ComputationForcefield"
+        key: str = ""
+        building_block: str = ""
+        coarse_grained_mapping: str = ""
+        implicit_solvent: str = ""
+        source: str = ""
+        description: str = ""
+        data: Union[Data, None] = None
+        citation: List[Citation] = field(default_factory=list)
+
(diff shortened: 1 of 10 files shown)


src/cript/nodes/subobjects/computation_forcefield.py
```diff
@@ -0,0 +1,127 @@
...
+        return self._json_attrs.data
+
+    @data.setter
+    def data(self, new_data: Union[Data, None]):
+        new_attrs = replace(self._json_attrs, data=new_data)
+        self._update_json_attrs_if_valid(new_attrs)
+
+    @property
+    def citation(self) -> List[Citation]:
+        return self._json_attrs.citation.copy()
```
nh916, line 122: why returning the copy instead of the attribute?

nh916 (approved):

src/cript/nodes/subobjects/computation_forcefield.py
```diff
@@ -0,0 +1,127 @@
...
+    class JsonAttributes(BaseNode.JsonAttributes):
+        node: str = "ComputationForcefield"
+        key: str = ""
+        building_block: str = ""
+        coarse_grained_mapping: str = ""
+        implicit_solvent: str = ""
+        source: str = ""
+        description: str = ""
+        data: Union[Data, None] = None
+        citation: List[Citation] = field(default_factory=list)
```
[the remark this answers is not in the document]
InnocentBug: The citation is a list, at least the data model says so.
And for lists we have been using empty lists as default. So that is what I am implementing here.

src/cript/nodes/subobjects/computation_forcefield.py
InnocentBug: citation is a list. and python returns a shallow copy of lists, so you would be able to change the members of the original list directly.
This is prevented by returning a copy of the list, so any modification on the copy doesn't change the node list.

For encapsulation of the list, this is necessary.

nh916: got it! You are 100% right, this should not have been a comment

CVilla17 (approved):

PR #49, opened 2023-03-30
Redesign of the API

InnocentBug:
# Description
We had some concerns with caching the API and schema and vocab.

This is a design suggestion. It is not conform with the design document, but I think it is incredibly elegant and powerful.
Not much can go wrong with this design.

## Changes

Vacab and Schema are both classes instead of raw dicts.
That hides the actual dicsts from the user, which is nice because then they can't mess it up.

Plus we have memeber function for validation instead of module function.
So more an object oriented design.

Both of these classes are automatically established inside the API object as it connects with the backend.
The data is then cached for the lifetime API object and constant.

It can be globally accessed via the globally cached API object.

## Known Issues
The tests don't work properly, because at the moment we can't fully integrate the back end.
We can't GET the entire vocab for example.

## Checklist


nh916:
I dislike this design because it makes understanding it and working with it much harder.

I don't see the need for vocabulary to be a class. Having it simply as 2 functions of `get_vocab` and `is_vocab_valid` is all that we need, as [implemented currently](https://github.com/C-Accel-CRIPT/Python-SDK/blob/develop/src/cript/api/vocabulary.py)

I don't see the need for schema validation to be a class. Having it simply as 2 functions of `get_schema` and `is_schema_valid` is all that we need, as [implemented currently](https://github.com/C-Accel-CRIPT/Python-SDK/blob/develop/src/cript/api/schema_validation.py)

I think there is some merit to making the api a context manager because then every thing that needs access to the `host`, such as `get_vocab` to get the controlled vocabulary and `get_db_schema` to get the db schema from the correct host, can all access the host of the api object and fetch what they need correctly.

If `db_schema_validation` and `is_vocab_valid` are both functions then they wouldn't need anything more than just the `host` of the api object. I subscribe to the principle of least privilege that runs code with the lowest privilege/permission level possible. I think if we allow anyone and everyone to access the api object then there is more of a chance that the api object can be misused, either from the user and bumping into errors and emailing us for support, or we could be giving anyone maintaining the SDK the ability to write spaghetti code that could make the SDK into a mess. We can prevent a lot of the issues ahead by complying with the principle of least privilege that is just best practice. Having the entire api object as a global variable just opens up a lot of issues that are not needed and can be happily solved with just functions.

We have to also keep in mind to not put the cart before the horse. As far as I understand the API currently does not have the ability to create a private instance and it might take some time before this is available. For us, that means that we might be able to push off private instances of the SDK for a bit and not focus on it too heavily. While it is good to plan ahead, this might be something that we do not have enough information yet to plan for and we might end up solving the wrong problem if we try to tackle it too early in my opinion. When the implementation is created there is a chance that it might look either exactly as we anticipate it or it might be very different than we had imagined.

I am not sure how the connect and disconnect method could work. I think a `connect` and `disconnect` method could be cool, but since we are using a token to interact with the backend I don't think it is necessary nor would it do much. If I call the `disconnect` method would it just replace the class variable token with an empty string `token =""` ? I can't imagine a lot of use cases for it either. 

I think returning a copy of the API object could create a bigger issue than is needed because then you are telling the user/maintainer that this is a private variable, but when they go to manipulate it, at their own risk, they either get an error that is puzzling to them or worse they never see any changes actually reflected, and be stuck in confusion as to why it is not working as they had expected. There are times that such code is needed out of necessity, but in this case I don't think this code is necessary and can be avoided easily. I don't think anything more than the `host` is needed from the api object. The pythonic way is that anything that does not need to be a class should be written as a function. The idea is that functions are often more lightweight and easier to read and maintain compared to classes, especially if the functionality being implemented doesn't require state management or complex behavior.

InnocentBug:
Let me explain exactly, why I designed this PR the way it is right now.

1. `get_vocab` is ill defined in the current implementation. The request of a vocab is only valid if you ask from where. You could absolutely add an argument to the `get_vocab` function that represents an API object.
Only with a host address it makes sense to request a vocab. (That argument can be optional with the meaning that it request the globally connected API instead.)
The current implementation is unclear in its meaning and error prone.
It only works if there is a globally defined API, so it requires users to create that object first. This concept is already confusing, because you would expect that you can call a module defined function any time.
Even worse is that we are caching that current vocab separately from the globally connected API. So in the event that API connection changes, the two caches become disconnected without a warning to the users leading to unexpected results.

2. `is_vocab_valid` is also ill defined in the current context. You cannot ask if a key value pair is valid with the vocabulary without specifying which vocabulary is you are checking against.
In the current context, it checks against the `get_vocab` vocab which suffers all the consequences from 1, plus this is not transparent to the user so it might lead to unexpected results. One solution (but not a great one would be to explicitly require the user to provide the vocab as an argument to this function.)
In order to make this a well defined request, it is logical (and the core principle of object oriented design) to combine data (here the vocab) with the functions that operate on the data in a class.

3. You might argue that we do not anticipate currently that users connect to different back ends and that is fair we do not.
First, even if we don't it is bad design to assume something without checking it. And that is what we currently do.
Plus, we already have the scenario of different back ends: the development and production back ends. And for testing having these two explicit is important.
Plus at some point, we need to move our CI testing against a self hosted docker container or something.
So we are connecting already against different back ends (although I admit not switching between them.)

4. `get_db_schema` suffers the same issues as described in 1.)

5. `is_schema_valid` suffers the same issues as described in 2.)

6. We already have an entire API object globally cached. That is implemented currently.
    And the `get_vocab`, `is_vocab_valid`, `get_db_schema`, and `is_schema_valid` make use of it.
    They wouldn't work without. If you don't give me an example of what exact issues could be caused by having a cached API object, I do not believe this argument.

7. Reasonable caching is already implemented via the context manager. `connect` and `disconnect` are the manual equivalent to the context manager to be used only with Jupyter Notebooks, since they don't handle context managers well.
This is also documented.
And the name `connect` and `disconnect` is a name for the user, not the devs. In reality they do not hold a connection with the back end right now, since we authenticate with the full token at the moment.
Right now they only handle the caching. But the names feel natural to connect and disconnect to a back end.
(Later we might want better hand shakes with the back end and this would be trivial to retro fit then.)

8. You would have to propose a better caching scheme for `host`, `db_schema`, and `vocab` to convince me of a different design. Also see 10.

9. I don't understand how this design promotes spaghetti code. It is much more object orientated than the current design.

10. The `API` represents the back end in the Python SDK. And logically the back end *has* a `db_schema` and *has* a `vocab`. So the object oriented realization is to make `db_schema` and `vocab` members of the `API` class. This is composition. And implemented with this design. 
https://stackoverflow.com/questions/2218937/has-a-is-a-terminology-in-object-oriented-language
This also makes it trivial to keep current connected back end and `db_schema` and `vocab` in sync. see 3.

11. Returning a copy or not a copy of the current global API is up for debate.
I don't feel as strongly about this one. I feel like it should be a copy, but can go otherwise as well.
At least it is all private.

nh916:
I hear you, and those are fair points.

Can we have the vocab and db schema as methods, and the cache as class variables within the API class instead of each one being a class?

`connect` and `disconnect` method, I don't love and would rather use `__enter__` and `__exit__` if we are planning on using it as a context manager, and have the cache clean up within the `__exit__`, but I can work with `connect` and `disconnect` as well.

Before going forward with this new design ideally I would want someone who is more senior to look at this design before we go ahead with it, but it might be the case that we are the only ones that are involved in this level detail. In that case then I can be okay with replacing our current implementation with the new API design

InnocentBug:
Yes, we could store the data directly in the API. I don’t love it, but we can. 
i can implement something like that.

InnocentBug:
Yes, connect and disconnect aren’t great. 

Everything should be context managers. And we still support that. 

However. Context managers don’t work great in Jupiter notebooks. So we offer these two for that situation

InnocentBug:
It’s fair that someone should look over the design. 
mid you find someone who is willing to. Great.

InnocentBug:
I implemented all your requests @nh916 

It is not my favorite design, but should be OK now.

If you find some one senior to look over this design, that would be great.

nh916:
> Yes, connect and disconnect aren’t great.
> 
> Everything should be context managers. And we still support that.
> 
> However. Context managers don’t work great in Jupiter notebooks. So we offer these two for that situation

okay I see, in that case, we could have the context manager call `connect` and `disconnect` under the hood. That could be a clean solution!

InnocentBug:
> > Yes, connect and disconnect aren’t great.
> > Everything should be context managers. And we still support that.
> > However. Context managers don’t work great in Jupiter notebooks. So we offer these two for that situation
> 
> okay I see, in that case, we could have the context manager call `connect` and `disconnect` under the hood. That could be a clean solution!

This is how it is implemented right now.

src/cript/api/api.py
```diff
@@ -100,18 +113,167 @@ def __init__(self, host: Union[str, None], token: [str, None]) -> None:
...
+        dict
+        controlled vocabulary
+        """
+
+        if vocab_category not in self._vocabulary:
+            raise InvalidVocabularyCategory(vocab_category, self._vocabulary.keys())
+
+        # Again, return a copy because we don't want
+        # anyone being able to change the private attribute
+        return copy.deepcopy(self._vocabulary[vocab_category])
```
nh916, line 149: Agreed!

nh916 (approved):

CVilla17 (approved):

PR #54, opened 2023-04-05
setup paginator.py

nh916:
# Description
Started setting up on paginator class. From previous conversations we said that this class would not be our first priority, and  we might push it back to future versions in the interest of time. I had some time so I went ahead and started setting it up with the blueprint that we had spoken about. Let me know your thoughts

### DIFF
--- a/src/cript/api/paginator.py
+++ b/src/cript/api/paginator.py
@@ -1,0 +1,48 @@
+from typing import Any, List
+
+
+class Paginator:
+    """
+    Paginator to flip through different pages of data that the API returns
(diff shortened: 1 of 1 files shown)


CVilla17 (dismissed):
Not really my expertise, but looks right. Should there be some thing to indicate total amount of pages?

nh916:
This PR is taking too long to be inputted into [wip_api](https://github.com/C-Accel-CRIPT/cript/pull/92) because it is needed there and this PR approval might take long to be approved and slow down development

PR #60, opened 2023-04-13
Create CI/CD of dependency security scanner

nh916:
# Description
Dependency security scanning. This could be a good security tool to scan any issues with the packages that we are using. This would ensure that our package is secure because even if our code is secure, but we are using a package or package version that is unsecure then it would create security vulnerabilities within our package.

This workflow comes directly from GitHub and is not by a third party


nh916:
@InnocentBug I tried formatting with trunk, but it seems like it is not working from my end, when you get a chance can you try formatting this with trunk

InnocentBug:
> @InnocentBug I tried formatting with trunk, but it seems like it is not working from my end, when you get a chance can you try formatting this with trunk

done

InnocentBug (approved):
Does that also support pyproject.toml?

nh916:
> Does that also support pyproject.toml?

I am uncertain to be honest

CVilla17 (approved):

PR #72, opened 2023-04-19
Introduce UID management to nodes and JSON

InnocentBug:
# Description
Introducing UID generation and management into the SDK.

UIDs are uuid4 strings with a "_" prefix.
We rely on the auto uuid no collision statement for this.

## Changes

Primary nodes and software and reference have a uid now.
This is auto generated at introducing the nodes.

Serializations is setup in a way that a nodes with uid is only serialized once, after that it is serialized via its uid.
Serialization to json adds a more controlled option with arguments that also returns the handled uids.
## Tests

Major overhaul of the subobject tests.
New structure in files and using more pytest fixtures.

## Known Issues

At the moment the serialization relies on the static class variables to track the already processed uids.
That isn't elegant but the best solution I see so far.

## Checklist


nh916:
I thought UID should start with `_:` no?

InnocentBug:
> I thought UID should start with `_:` no?

I don't know. You know better then I.
I changed it to `_:` now

nh916 (dismissed):

nh916:
I'll add @brili to this as well, just in case we are doing something wrong

InnocentBug:
I updated the UID serialization now, such that a previously serialized node is not just a string, but a node with only `node` and `uid` attribute.

Also added a deepcopy functionality that changes the uid upon calling deepcopy of a node.
Later we will have to add url and uuid management to the deepcopy.

I know this looks a bit bigger, because I refactored the tests, but that was overdue anyways.

nh916:
> At the moment the serialization relies on the static class variables to track the already processed uids.
That isn't elegant but the best solution I see so far.

Can we use a hashmap or hashset to keep track of visited nodes UID or am I misunderstanding this?

tests/conftest.py
```diff
@@ -307,85 +290,281 @@ def simple_inventory_node() -> None:
...
+        measurement_id=2,
+        material=[simple_material_node],
+        data=simple_data_node,
+    )
+    return c
+
+
+@pytest.fixture(scope="function")
+def simple_condition_dict(simple_material_node, simple_data_node) -> dict:
+    ret_dict = {
```
nh916, line 472: I think `condition_dict` would have been a more self documenting name here

tests/conftest.py
```diff
@@ -307,85 +290,281 @@ def simple_inventory_node() -> None:
...
 
-    return my_citation
+@pytest.fixture(scope="function")
+def simple_citation_dict(simple_reference_dict) -> dict:
+    ret_dict = {"node": ["Citation"], "reference": simple_reference_dict, "type": "reference"}
+    return ret_dict
 
 
 @pytest.fixture(scope="function")
 def simple_quantity_node() -> cript.Quantity:
```
nh916, line 388: I think it might be a good idea to get some docstrings for these functions

tests/conftest.py
```diff
@@ -307,85 +290,281 @@ def simple_inventory_node() -> None:
...
-    minimal quantity node
-    """
-    my_quantity = cript.Quantity(key="mass", value=1.23, unit="gram")
+    quantity = cript.Quantity("mass", 11.2, "kg", 0.2, "std")
+    return quantity
 
-    return my_quantity
+
+@pytest.fixture(scope="function")
+def simple_quantity_dict() -> dict:
```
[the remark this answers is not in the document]
nh916: oh I see so you're making the node fixture and then you're making the dict that would be if it were to be serialized.

thats not a bad idea, so updating the fixture and the dict would be easier and more maintainable, but I would want to start separating out the files at some point to keep primary nodes in one file, dicts in another, and so on because otherwise we are just throwing everything into one file and the file will become unruly

nh916:
at some point we should split up our `conftest.py` into:
* `fixtures/primary_nodes/`
* `fixtures/subobjects/`
* `fixtures/supporting_nodes/`

and maybe it might be a good idea to have 
* `fixtures/primary_nodes/simple_nodes,py`
* `fixtures/primary_nodes/simple_nodes_dict,py`

so that our conftest.py doesn't get unruly and become too big

tests/conftest.py
```diff
@@ -307,85 +290,281 @@ def simple_inventory_node() -> None:
...
+
+@pytest.fixture(scope="function")
+def simple_software_node() -> cript.Software:
+    software = cript.Software("SOMA", "0.7.0", "https://gitlab.com/InnocentBug/SOMA")
+    return software
+
+
+@pytest.fixture(scope="function")
+def simple_software_dict() -> dict:
+    ret_dict = {"node": ["Software"], "name": "SOMA", "version": "0.7.0", "source": "https://gitlab.com/InnocentBug/SOMA"}
```
nh916, line 407: `software_dict` would have been a better variable name instead of `ret_dict`

tests/conftest.py
```diff
@@ -307,85 +290,281 @@ def simple_inventory_node() -> None:
...
+    return ret_dict
+
+
+@pytest.fixture(scope="function")
+def simple_property_node(simple_material_node, simple_condition_node, simple_citation_node, simple_data_node, simple_process_node, simple_computation_node):
+    p = cript.Property(
```
nh916, line 413: `p` is a very bad variable name, variable names should be self documenting

tests/conftest.py
```diff
@@ -307,85 +290,281 @@ def simple_inventory_node() -> None:
...
+
+
+@pytest.fixture(scope="function")
+def simple_software_dict() -> dict:
+    ret_dict = {"node": ["Software"], "name": "SOMA", "version": "0.7.0", "source": "https://gitlab.com/InnocentBug/SOMA"}
+    return ret_dict
+
+
+@pytest.fixture(scope="function")
+def simple_property_node(simple_material_node, simple_condition_node, simple_citation_node, simple_data_node, simple_process_node, simple_computation_node):
```
nh916, line 412: this fixtuer says that it is a `simple property node` meaning that it has the bare required arguments and is the smallest node that it can be, however, this node has all the possible arguments for it, this seems more like `complex_property_node`


my naming convention for fixtures so far has been:
* `simple_node`: are nodes with just the required arguments and nothing more keeping them very simple
* `complex_nodes`: are nodes with all possible required arguments making them as big as possible

tests/nodes/primary_nodes/test_experiment.py
```diff
@@ -169,23 +171,14 @@ def test_experiment_json(simple_process_node, simple_computation_node, simple_co
...
-                "node": ["Citation"],
-                "type": "derived_from",
-                "reference": {"node": ["Reference"], "type": "journal_article", "title": "'Living' Polymers"},
-            }
-        ],
+        "citation": [simple_citation_dict],
     }
 
-    print("\n\n")
-    print("----------------------------------------------------")
```
nh916, line 182: thank you for stripping out the print statements. big help!

tests/nodes/subobjects/test_computation_forcefiled.py
```diff
@@ -0,0 +1,39 @@
+import json
+
+from util import strip_uid_from_dict
+
+import cript
+
+
+def test_computation_forcefield(simple_computation_forcefield_node, simple_computation_forcefield_dict):
+    cf = simple_computation_forcefield_node
```
nh916, line 9: variable names should be self documenting

tests/nodes/subobjects/test_computation_forcefiled.py
```diff
@@ -0,0 +1,39 @@
+import json
+
+from util import strip_uid_from_dict
+
+import cript
+
+
+def test_computation_forcefield(simple_computation_forcefield_node, simple_computation_forcefield_dict):
```
nh916, line 8: it might be a good idea to put a small docstring in each test to say what it is testing

tests/util.py
```diff
@@ -0,0 +1,21 @@
+import copy
+
+
+def strip_uid_from_dict(node_dict):
+    """
+    Remove "uid" attributes from nested dictionaries.
```
nh916, line 6: I need more explanation to understand what you're testing here

nh916 (approved):

InnocentBug:
> > At the moment the serialization relies on the static class variables to track the already processed uids.
> > That isn't elegant but the best solution I see so far.
> 
> Can we use a hashmap or hashset to keep track of visited nodes UID or am I misunderstanding this?

I am using a `set` for this. So object have to be hasable (we are tracking the uid strings, so that is no problem.)
So I am assuming the `in` python operator does that with a hash based search.

The less elegant part is, that the `json` Serialization class that I am using is not meant to do this.

InnocentBug:
> at some point we should split up our `conftest.py` into:
> 
>     * `fixtures/primary_nodes/`
> 
>     * `fixtures/subobjects/`
> 
>     * `fixtures/supporting_nodes/`
> 
> 
> and maybe it might be a good idea to have
> 
>     * `fixtures/primary_nodes/simple_nodes,py`
> 
>     * `fixtures/primary_nodes/simple_nodes_dict,py`
> 
> 
> so that our conftest.py doesn't get unruly and become too big

Kind of done.
I didn't seperate dicts and nodes though.
I think it more naturally to maintain them at the same place.

tests/conftest.py
```diff
@@ -307,85 +290,281 @@ def simple_inventory_node() -> None:
...
+
+@pytest.fixture(scope="function")
+def simple_quantity_dict() -> dict:
+    ret_dict = {"node": ["Quantity"], "key": "mass", "value": 11.2, "unit": "kg", "uncertainty": 0.2, "uncertainty_type": "std"}
+    return ret_dict
+
+
+@pytest.fixture(scope="function")
+def simple_software_node() -> cript.Software:
+    software = cript.Software("SOMA", "0.7.0", "https://gitlab.com/InnocentBug/SOMA")
```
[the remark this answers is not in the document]
InnocentBug: possible, but that are actually two different code paths.
I use it with arguments, yours are keyword arguments.
So for testing, strictly speaking we should do both.

So I am a little reluctant to remove it here.

tests/conftest.py
InnocentBug: This is testing, so I don't care as much about readability and documentation.
Since this is inside a fixture with a telling name, I am leaning towards readable enough here.

tests/conftest.py
InnocentBug: I don't know what to write in those to be hones.
They are very short functions with only one active line of code each.
If you want to write some, sure go ahead.

tests/conftest.py
InnocentBug: same as above. I don't care as much in tests and it is a two line functions.

tests/conftest.py
InnocentBug: I know, but I don't care enough for tests to fix it.

InnocentBug: If you actually wanted to prevent this, you can set up pylint with trunk to sort out for bad variable names.

tests/conftest.py
InnocentBug: I didn't not know your nomenclature.
I can rename them.
Actually all of my fixtures use as many attributes as possible, because that way I cover more of the code.

nh916: yeah I was doing that at first too, but I think somewhere along the testing I came to the conclusion that if I am putting everything it is harder to test what is failing and trace the source of the bug, so I started doing both small tests and big tests where I thought it was clear to me. Let me know if I am not making any sense. 
I could also be wrong as well, feel free to let me know

I should probably document the `simple` and `complex` node convention somewhere so it is clear because it probably wouldn't be as clear for a new contributor just joining

InnocentBug: I renamed the nodes accordingly.
There is a lot more complex more then before.

tests/util.py
InnocentBug: What are you missing.
The doc says, that I am removing all uid attributes.

uids are easily different, so for test purposes sometimes we want to make sure that we can ignore this.
Hence this helper funciton.

tests/conftest.py
nh916: > > at some point we should split up our `conftest.py` into:
> > ```
> > * `fixtures/primary_nodes/`
> > 
> > * `fixtures/subobjects/`
> > 
> > * `fixtures/supporting_nodes/`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and maybe it might be a good idea to have
> > ```
> > * `fixtures/primary_nodes/simple_nodes,py`
> > 
> > * `fixtures/primary_nodes/simple_nodes_dict,py`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so that our conftest.py doesn't get unruly and become too big
> 
> Kind of done. I didn't seperate dicts and nodes though. I think it more naturally to maintain them at the same place.

NICE!!!

I did not see that, but I just checked it out and saw that you are already ahead of the game! 🎉 🎉  🎉

nh916 (approved):

CVilla17 (approved):

PR #74, opened 2023-04-24
Use class name for `node` attribute

InnocentBug:
# Description

Small simplification in the `node` assignment process.
Guarantees that class name and `node` attribute are identical.

## Changes

It also offers the option to get the `node_type` from classes as well as instances.

```python
import cript
assert cript.Material.node_type == "Material"
mat = cript.Material("asfd", "asdf")
assert mat.node_type == "Material"
```

## Known Issues
If we have a different capitilazation strategy between class names and node attributes, we need to add one little extra step.

## Checklist


nh916 (approved):

nh916 (approved):

nh916:
merging to `wip_api`

PR #77, opened 2023-04-27
move save to project node

InnocentBug:
# Description
Move the save function to the project node instead of API.
## Changes
Transfer the memebership of who can save.
This way it is much clearer, that you can only save entire projects.
It also is transparent to the user which api is used (like before).

No check, that the passed node to API is actually a project and not something else.

## Tests
adjusted tests to reflect this change
## Checklist


InnocentBug:
We may consider this at a later time.

PR #78, opened 2023-04-28
Validate nodes

InnocentBug:
# Description
Validated all nodes against db schema

## Changes
All the changes, that are required to be compatible with the db schema
* removed process from material node
* `added simple_computational_node fixture`
* simplified fixtures by using other fixtures instead of rewriting the nodes for the fixtures
* made node attributes singular as API requires them
* changed node vocabularies to match the db schema enums
    * mainly material identifier only uses `bigsmiles` instead of other identifiers
* reduced complexity of some tests to make maintenance easier
   * for example, instead of having 2 process product, there is only one
* changing uncertainty from int to str as db schema requires
* added `_is_node_field_valid` function that can be used within the code to keep all the node field validations in there and make the code more DRY
* removing some instances of `copy.copy()` where I don't see it being valid

## Notes
* added simple fixtures for subobjects instead of using complex nodes
   * simple nodes make it easier to debug and work with

## Checklist


nh916:
#  This PR must be looked through very thoroughly

because a lot has been changed and renamed and if something turns out to be wrong, it can be a headache to fix again and then clean up.

src/cript/nodes/core.py
```diff
@@ -98,9 +122,12 @@ def _from_json(cls, json_dict: dict):
         # The call to the constructor might ignore fields that are usually not writable.
         try:
             node = cls(**arguments)
+        # TODO we should not catch all exceptions if we are handling them, and instead let it fail
```
InnocentBug, line 125: We are not catching the exception here.
We are reraising it.
See 129.
But before we are printing some extra info that can help debugging.

It's not ideal, but it isn't catching the exception.

nh916:
I think it might be nicer if we keep the `__init__.py` empty, but have files in `src/cript/nodes/utils/`

src/cript/nodes/exceptions.py
```diff
@@ -61,6 +61,38 @@ def __str__(self):
         return f"JSON Serialization failed for node type {self.node_type} with JSON dict: {self.json_str}"
 
 
+class CRIPTAttributeModificationError(CRIPTException):
```
nh916, line 64: I would replace this with `UneditableAttributeError`

nh916:
I'm thinking of doing what you suggested before in this branch of merging everything into it, fixing as many of the tests that we can, and when stuck we comment out that test and leave a `TODO` and then merge it into develop.

nh916:
I was looking into a bit, but [pydantic](https://docs.pydantic.dev/dev-v2/) might be very helpful in serialization and deserialization. I have not looked too much into it yet, but it seemed promising. I think it could even give us the ability to easily rename things as we want, like instead of doing flattening of identifiers that we did before we could make it a bit simpler

```python
import json
from pydantic import BaseModel

class MyModel(BaseModel):
    data: str
    nested_list: List[List[int]]

    def serialize(self):
        serialized = {
            'data': self.data,
            'list': [list(sublist) for sublist in self.nested_list]
        }
        return json.dumps(serialized)

# Usage
my_model = MyModel(data='Hello', nested_list=[[1, 2], [3, 4]])
serialized_data = my_model.serialize()
print(serialized_data)
```

also we could give attributes `aliases` so something like `type` or `property` we could have in our code as `type_`, but in json form it comes out as `type`

```python
import json
from pydantic import BaseModel, Field

class MyModel(BaseModel):
    my_name: str = Field(..., alias='name')

my_model = MyModel(my_name='John')
serialized_data = my_model.json()
print(serialized_data)
```

```jon
{
  "name": "John"
}
```

src/cript/nodes/core.py
```diff
@@ -91,20 +147,33 @@ def _from_json(cls, json_dict: dict):
...
             raise exc
-        attrs = cls.JsonAttributes(**arguments)
-        # Handle UID manually. Conserve newly assigned uid if uid is default (empty)
-        if attrs.uid == cls.JsonAttributes().uid:
-            attrs = replace(attrs, uid=node.uid)
+
+        try:
+            attrs = cls.JsonAttributes(**arguments)
+        except TypeError:
+            help(cls.JsonAttributes)
```
nh916, line 171: do we need to have this here or was this left by mistake?

src/cript/nodes/exceptions.py
nh916: We can come back and iterate on this later

nh916 (approved):

src/cript/nodes/core.py
InnocentBug: good catch, this has to go.

nh916:
@InnocentBug this is a great achievement and a huge task you accomplished!!!

well done!!

nh916 (approved):

PR #82, opened 2023-05-01
Check for orphraned nodes

InnocentBug:
# Description
Validate project nodes, that they don't contain orphan nodes that should be listed in project or experiment.

## Changes
Extra validation of project graphs.
## Tests
Added tests to ensure logic is working.

## Known Issues

Fixes an additional issue of `inventory` by renaming `materials_list` to `materials`.
It prevented deep copies and json deserialization.

## Notes
Added a helper function that allows users to easily fix their graphs.

## Checklist

### DIFF
--- a/src/cript/nodes/exceptions.py
+++ b/src/cript/nodes/exceptions.py
@@ -1,6 +1,11 @@
 from abc import ABC, abstractmethod
 
 from cript.exceptions import CRIPTException
+from cript.nodes.primary_nodes.computation import Computation
+from cript.nodes.primary_nodes.computational_process import ComputationalProcess
+from cript.nodes.primary_nodes.data import Data
+from cript.nodes.primary_nodes.material import Material
+from cript.nodes.primary_nodes.process import Process
 
 
 class CRIPTNodeSchemaError(CRIPTException):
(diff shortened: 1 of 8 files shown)


src/cript/nodes/exceptions.py
```diff
@@ -57,3 +59,37 @@ def __init__(self, node_type: str, json_dict: str):
...
+class CRIPTOrphranedMaterialError(CRIPTOrphranedNodesError):
+    def __init__(self, orphaned_node: "Material"):
+        super().__init__(orphaned_node)
+
+    def __str__(self):
+        ret_string = "While validating a project graph, an orphraned material node was found. "
+        ret_string += "This material is present in the graph, but not listed in the project. "
+        ret_string += "Please add the node like: `your_project.materials += [orphaned_material]`. "
+        ret_string += f"The orphraned material was {self.orphraned_node}."
+        return ret_string
```
nh916, line 82: consider using this instead to make it cleaner

```python
def __str__(self) -> str:
    return (
        f"While validating a project graph, an orphaned material node was found. "
        f"This material is present in the graph, but not listed in the project. "
        f"Please add the node like: `your_project.materials += [orphaned_material]`. "
        f"The orphaned material was {self.orphraned_node}"
    )
```

src/cript/nodes/exceptions.py
```diff
@@ -57,3 +59,37 @@ def __init__(self, node_type: str, json_dict: str):
...
+
+
+class CRIPTOrphranedExperimentError(CRIPTOrphranedNodesError):
```
nh916, line 85: can we make one error `OrphanedNodeError` and just change the messages depending on how it is called?

I think that could make lives easier and we don't have to keep adding more exceptions for every new orphaned node, we just have one and we make it into a template and depending on how its called it gives the correct context for it

src/cript/nodes/primary_nodes/project.py
```diff
@@ -76,6 +80,37 @@ def __init__(
...
         self.validate()
 
+    def validate(self):
+        # First validate like other nodes
+        super().validate()
+
+        # Check graph for orphraned nodes, that should be listed in project
+        # Project.materials should contain all material nodes
+        all_materials = self.find_children({"node": ["Material"]})
+        for material in all_materials:
```
nh916, line 90: very picky small improvement, I would just name it as `for material in project_materials:`

src/cript/nodes/exceptions.py
```diff
@@ -57,3 +59,37 @@ def __init__(self, node_type: str, json_dict: str):
...
+    def __init__(self, orphaned_node):
+        super().__init__(orphaned_node)
+
+    def __str__(self):
+        node_name = self.orphaned_node.node_type.lower()
+        ret_string = f"While validating a project graph, an orphraned {node_name} node was found. "
+        ret_string += f"This {node_name} node is present in the graph, but not listed in any experiment of the  project. "
+        ret_string += f"Please add the node like: `your_experiment.{node_name} += [orphaned_{node_name}]`. "
+        ret_string += f"The orphraned {node_name} was {self.orphraned_node}."
+        return ret_string
```
nh916, line 95: I would change the error to

```python

class OrphranedNodeError:
    parent_node = None
    child_node = None

    def __init__(self, parent_node, child_node):
        self.parent_node = parent_node
        self.child_node = child_node

    def __str__(self) -> str:
    return (
        f"While validating the {self.parent_node.node_type} graph, an orphaned node: {self.child_node} was found. "
       f"The {self.child_node} node is present in the graph, but not listed in any {self.parent_node.node_type} "
       f"Please add the {self.child_node} to the appropriate {self.parent_node.node_type} "
       f"See documentation for more details"
    )
```

and then within the documentation we say 

```text
This error is raised when a child node is not attached to the 
appropriate parent node. For example, all material nodes used 
within a project must belong to the project inventory. If there is 
a material node that is used within a project but not a part of the 
inventory and the validation code finds it then it raises an `OrphranedNodeError`

Fixing this is simple and easy, just take the node that CRIPT Python SDK 
found a problem with and associate it with the appropriate parent via

my_project.inventory = my_orphaned_material_node
```

src/cript/nodes/primary_nodes/project.py
```diff
@@ -76,6 +80,37 @@ def __init__(
...
+            if material not in self.materials:
+                raise CRIPTOrphranedMaterialError(material)
+
+        # Check graph for orphraned nodes, that should be listed in the experiments
+        all_experiments = self.find_children({"node": ["Experiment"]})
```
nh916, line 95: small picky preference, but I would change this into something like

```python
all_experiments = self.find_children(node="Experiment")
```
I think thats just nicer and cleaner, and under the hood find_children puts the experiment within brackets and gets it prepared and searches for it

nh916:
@InnocentBug does this code work so far?
the logic within `validate` is a bit hard for me to follow and it feels a bit confusing, I dont know if thats just me or not

src/cript/nodes/primary_nodes/project.py
```diff
@@ -76,6 +80,37 @@ def __init__(
...
+        # All in the graph has to be in at least one experiment
+        node_types = ("Process", "Computation", "ComputationalProcess", "Data")
+        for node_type in node_types:
+            all_nodes = self.find_children({"node": [node_type]})
```
nh916, line 99: does this make any sense?

I think this is saying `within process find process`

src/cript/nodes/primary_nodes/project.py
```diff
@@ -76,6 +80,37 @@ def __init__(
...
+                node_type_attr = "computational_process"
+
+            for node in all_nodes:
+                node_found = False
+                for experiment in all_experiments:
+                    if node in getattr(experiment, node_type_attr):
+                        node_found = True
+                        break
+                if not node_found:
+                    raise CRIPTOrphranedExperimentError(node)
```
nh916, line 112: I'm struggling to understand this, is this a popular algorithm or something we created?

src/cript/nodes/exceptions.py
InnocentBug: No, on the contrary.
I actually want to diversify this error more.

We still have a common base node (and I will put most of the logic there), but different exceptions help a lot in fixing the error.

Consider this.

```python
active_experiment = project.find_children({"node": ["Experiment"]})[0]
try:
   project.validate()
except CRIPTOrphanedMaterialError as exc:
  project.materials += exc.orphaned_node
except CRIPTOrphanedDataError as exc:
  active_experiment.data += exc.orphaned_node
except CRIPTOrphanedProcessError as exc:
  active_experiment.process += exc.orphaned_node
  ```
  
  The second and thrid exception handling block is only possible if you can easily deduce the orphaned node type from the exception.

File: src/cript/nodes/primary_nodes/project.py (hunk 3 above)
Comment (InnocentBug) [reply src/cript/nodes/primary_nodes/project.py:90]: sure, np.

File: src/cript/nodes/exceptions.py (hunk 4 above)
Comment (InnocentBug) [reply src/cript/nodes/exceptions.py:95]: Sure, we can add the parent node and documentation string.

Comment (InnocentBug) [reply src/cript/nodes/exceptions.py:95]: I don't understand how the `parent_node` dynamics should work.
The parent is always of type project, so why have it dynamic?
Unless you thought if would be possible to to that for experiment, but it is actually not possible for experiment, because we don't know which experiment of the project the orphaned nodes belong to.

Comment (nh916) [reply src/cript/nodes/exceptions.py:95]: It was just something rough I was thinking of, if we can't have parent node then we could just have that hard coded and only pass in the child node. I'm happy either way.

I'm happy to brainstorm together as well and think of a good way to approach this!

File: src/cript/nodes/exceptions.py (hunk 2 above)
Comment (InnocentBug) [reply src/cript/nodes/exceptions.py]: see at the helper function, that I added. It actively uses this.

File: src/cript/nodes/primary_nodes/project.py (hunk 7 above)
Comment (InnocentBug) [reply src/cript/nodes/primary_nodes/project.py]: I simplified the algorithm.
(Instead of iterating lists, I compact the lists first and search in the gigantic list.)

File: src/cript/nodes/primary_nodes/project.py (hunk 5 above)
Comment (InnocentBug) [reply src/cript/nodes/primary_nodes/project.py]: Unfortunately, this doesn't work.

This is just a simplified use of `find_children` it can do much more.
For example with `{"node": ["Experiment"], "name": "my name"}` you can find an experiment with a name too.
this also works for nested things.

It might be able to do this with `kwargs` but I don't want to open this in this PR.
Maybe a separate PR can address this.

**Comment by InnocentBug** [comment]
> @InnocentBug does this code work so far? the logic within `validate` is a bit hard for me to follow and it feels a bit confusing, I dont know if thats just me or not

Code works great now.
I added extensive tests, that produces the error and fixes the errors too.

File: src/cript/nodes/exceptions.py [hunk 8]
```diff
@@ -57,3 +59,138 @@ def __init__(self, node_type: str, json_dict: str):
 
     def __str__(self):
         return f"JSON Serialization failed for node type {self.node_type} with JSON dict: {self.json_str}"
+
+
+class CRIPTOrphanedNodesError(CRIPTException, ABC):
```
nh916, line 64: possible idea, we could maybe in the future put all orphaned nodes exceptions into one file to keep to keep it cleaner, something like this

`src/cript/nodes/exceptions/exception.py`
`src/cript/nodes/exceptions/orphaned_nodes_exception.py`

src/cript/nodes/primary_nodes/project.py
nh916: yeah I agree, this is a separate PR thing, but just wanted to make a small suggestion for future. 

It might be a good idea to not have the arguments be a dictionary and instead be explicit arguments and as simple as possible like:

```python
all_experiments = self.find_children(node="Experiment", name="my_name", )
```

I just really dislike `**kwargs` because its less explicit and if I am the user then I then have to go to documentation to see what is allowed and see how I can get clever about it to make it do what I need and if I missed anything. dictionary arguments are very open ended in my opinion, whereas explicit arguments you can't mess it up, and everything is right there and obvious . 

Unless there are too many possibilities where the function would have to have a million arguments to capture all possible arguments and situations that the function can be used in, and then a `**kwargs`  argument makes more sense.

I'm getting too far into it, but just wanted to put some food for thought and some ideas for later and we can come back to this later if we need to as well

src/cript/nodes/primary_nodes/project.py
```diff
@@ -76,6 +76,50 @@ def __init__(
...
+        super().validate()
+
+        # Check graph for orphaned nodes, that should be listed in project
+        # Project.materials should contain all material nodes
+        project_graph_materials = self.find_children({"node": ["Material"]})
+        # Combine all materials listed in the project inventories
+        project_inventory_materials = []
+        for inventory in self.find_children({"node": ["Inventory"]}):
+            for material in inventory.materials:
+                project_inventory_materials.append(material)
```
nh916, line 95: this makes our code more implicit, do we want to make this more explicit?

I don't know enough on how to design this part, I think we need someone more senior than me who knows better

tests/test_node_util.py
```diff
@@ -114,3 +122,80 @@ def raise_node_dict(node_dict):
...
+    project.collections[0].inventories += [cript.Inventory("test_inventory", materials=[material])]
+    project.validate()
+    # Reverse this fix
+    project.collections[0].inventories = []
+    with pytest.raises(CRIPTOrphanedMaterialError):
+        project.validate()
+
+    # Fix by add to the materials list instead.
+    # Using the util helper function for this.
+    cript.add_orphaned_nodes_to_project(project, active_experiment=None)
```
nh916, line 151: this could be a good idea, we don't implicitly fix it for them, but we give them a utility that they can fix it for themselves. 

I dont know, I think someone more senior than me would probably know the best design for this and how to go about it, it's a bit outside my depth of knowledge

nh916 (approved):

src/cript/nodes/primary_nodes/project.py
nh916: I think that would probably improve readability but not do much in terms of performance, but not sure to be honest

InnocentBug: yeah, this is not optimized for performance at all

src/cript/nodes/exceptions.py
InnocentBug: sure, we can do that.
Not sure we need it now yet.

src/cript/nodes/primary_nodes/project.py
InnocentBug: I don't know.
Maybe. it's just a double list concat.

PR #85, opened 2023-05-02
Fix API Exceptions

nh916:
# Description

Made the CRIPT Python SDK exceptions more user friendly and added documentation for them

## Changes
* 

## Tests
* All the tests are running fine
* The API Exceptions documentation is generating fine
image

## Known Issues
* documentation UI does not look very beautiful and is a bit unpleasant to look through
   * this is a problem that should be improved/upgraded in the future
* `APIError` is a bit generic and could in the future be broken down into smaller and more specific exceptions

## Checklist


nh916:
I think if I remove the `__init__` and `__str__` docstrings they will g away within the documentation, which is what I think we want in this scenario

InnocentBug (dismissed):

nh916:
@InnocentBug what are your thoughts on making the Exceptions documentation like this as opposed to before?

this way the exceptions methods don't have any docstrings, but I think that's okay because they're all very simple. what are your thoughts?

image

nh916:
I changed it a bit to make it cleaner and more consistent, I think this version is much better

CVilla17 (approved):

PR #87, opened 2023-05-02
Fix node exceptions

nh916:
# Description
Wrote documentation for Node Exceptions

## Changes
* added type hinting to methods and messages
* formatted error messages
* updated some packages
    * > the updates I believe already exist in the `develop` branch but might not yet exist in `validate` branch

## Known Issues
*  `CRIPTJsonDeserializationError` docs needs updates down the line
    * more documentation is needed on how to troubleshoot this section
    * better explanation is needed on how and why this error occurs as well as an example of where it would and wouldn't occur
* in order to get the documentation clean with only `Definition` and `How to fix` section, the rest of the method docstrings do not exist
    * I think this is okay as the methods are very simple and do not need much explanation, and simple comments I think might be enough

## Notes
* Some of these errors are hard to troubleshoot and understand, lets see if we can update and fix them a bit
* We could probably shorten the error names a bit by removing the word `CRIPT` when not needed. 
    * I'm thinking they are using our package so of course they know that its our errors, so we can remove it from the name
    * We could only add `CRIPT` when other packages/modules have the same error name 
         * When I was first creating the `ConnectionError` exception I noticed that 
         [requests](https://requests.readthedocs.io/en/latest/) library has the same exact error name that [requests library had an exception](https://requests.readthedocs.io/projects/cn/zh_CN/latest/_modules/requests/exceptions.html?highlight=connection%20error) named the same thing, 
so I renamed ours to `CRIPTConnectionError` I think in those instances it makes sense, but in other instances we can just remove it

## Screenshot
image


src/cript/api/api.py
```diff
@@ -417,26 +417,25 @@ def _is_node_schema_valid(self, node_json: str) -> bool:
 
         db_schema = self._get_db_schema()
 
-        node_dict = json.loads(node_json)
-        try:
```
InnocentBug, line 421: is it intended to not throw a custom exception here anymore?

src/cript/api/api.py
```diff
@@ -417,26 +417,25 @@ def _is_node_schema_valid(self, node_json: str) -> bool:
...
         else:
-            raise CRIPTJsonNodeError(node_list, str(node_list))
+            raise CRIPTJsonNodeError(node_type_list, str(node_type_list))
 
         # set which node you are using schema validation for
         db_schema["$ref"] = f"#/$defs/{node_type}Post"
 
         try:
             jsonschema.validate(instance=node_dict, schema=db_schema)
-        except jsonschema.exceptions.ValidationError as error:
```
InnocentBug, line 438: This error message is superior to the new one.

InnocentBug: at least raise the exception from the old one.

InnocentBug (changes requested):
This seems out of sync.

Some changes revert what we added and some exceptions are just missing.
Seems like a rebase from something old, can you check that?

For example the orphan exception are removed for some reason.

nh916:
closing this PR in favor of the [updated one](https://github.com/C-Accel-CRIPT/Python-SDK/pull/110)

PR #88, opened 2023-05-02
successfully flatten the identifiers

nh916:
# Description
flattens out the identifiers list from a list of dicts to be on the dict itself

## Changes
made a change to the node encoder. 

Before the serialization is fully done I check if the node is a "Material" node and if it is then I send it to 
`material_identifiers_list_to_json_fields` function to flatten the identifiers into the serialized_dict
and set the serialized_dict to the output of `material_identifiers_list_to_json_fields` 

## Tests
* This serialization can be posted to the API
* 

## Known Issues
* function name to flatten identifiers is very long and should be shortened to something easily understandable
* comments and docstrings could be cleaner

## Notes
Have not updated tests to work with the new code, but I can do that if we like it

### DIFF
--- a/src/cript/nodes/util.py
+++ b/src/cript/nodes/util.py
@@ -38,8 +38,59 @@ class NodeEncoder(json.JSONEncoder):
                     if getattr(obj._json_attrs, key) != default_values[key]:
                         serialize_dict[key] = getattr(obj._json_attrs, key)
             serialize_dict["node"] = obj._json_attrs.node
+
+            # if node is material, then convert the identifiers list to JSON fields
+            if serialize_dict["node"] == ["Material"]:
+                serialize_dict = material_identifiers_list_to_json_fields(serialize_dict)
+
             return serialize_dict
         return json.JSONEncoder.default(self, obj)
+
+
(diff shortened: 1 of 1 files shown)


nh916:
@InnocentBug I think this should be a clean design and the rest of the nodes should be easy to change to get it to work with the API. what do you think?

InnocentBug:
Hmm, this is one approach to this.
Unfortunately, I have already put the work in to use a different approach.
https://github.com/C-Accel-CRIPT/Python-SDK/pull/78/commits/62be24d23b27119504115c2fdbfa8d0610816d97
as part of #78

Not sure what to do now.

nh916:
> Hmm, this is one approach to this. Unfortunately, I have already put the work in to use a different approach. [62be24d](https://github.com/C-Accel-CRIPT/Python-SDK/commit/62be24d23b27119504115c2fdbfa8d0610816d97) as part of #78
> 
> Not sure what to do now.

oh is this where we put every single identifier field on the SDK as an attribute?

nh916:
we already have this implemented and deserialization in the validate branch

PR #89, opened 2023-05-03
added documentation for cript.utils

nh916:
# Description
set up documentation for utility functions of the CRIPT Python SDK

## Changes
* added `Utility Functions` documentation
    * added it to mkdocs.yaml
    * created an md file for it
* wrote docstrings for `load_nodes_from_json`

## Screenshot
image


nh916:
@InnocentBug what are your thoughts on doing a schema validation of the node before converting it from JSON to Python node?

nh916:
I think it may be a better idea to hold off on this for now until because we are working heavily on utils functions and there might be a lot of changes and updates that would make this outdated

InnocentBug:
> @InnocentBug what are your thoughts on doing a schema validation of the node before converting it from JSON to Python node?

yeah, we should do that.

nh916:
@InnocentBug wrote this a while back and I was thinking of holding off until later, but I'm thinking of just putting it into the `validate-branch` for now. Not sure if I should hold off or not.

I think if we keep putting things into the validate branch it might end up becoming giant and at the end create a log merge conflicts, but I am not sure if we should merge it into develop either because some of the serialization and deserialization code needs clean up. what are your thoughts?

InnocentBug:
> @InnocentBug what are your thoughts on doing a schema validation of the node before converting it from JSON to Python node?

Would be great, but doesn't work.
A JSON from the backend does not conform with the schema.

InnocentBug (approved):

PR #91, opened 2023-05-04
UUID support for all nodes that need it

InnocentBug:
# Description
This adds UUIDs to all nodes that need one.

## Known Issues

Not all the way there for serialization, but one step in this direction

## Notes

merge into validate nodes, since it doesn't break anything there and seems nice to have there.

## Checklist


nh916:
@InnocentBug I have a bunch of files that are not merged yet and I am afraid this would create a lot of merge conflicts and we would lose some code when trying to resolve the merges.

the only thing that I can think of is waiting until I get all the tests to pass and then try to merge these branches into each other, what are your thoughts?

nh916:
The SDK is very vulnerable right now and can break easily and badly right now. we have to treat it very carefully right now until we can get everything renamed and working again. We have to coordinate any changes between each other very carefully until we can get everything renamed and working again

src/cript/api/api.py
```diff
@@ -27,6 +27,8 @@
 # Use the `_get_global_cached_api for access.
 _global_cached_api = None
 
+_API_VERSION_STRING = "v1"
```
nh916, line 30: I suggest putting this within `_prepare_host()` function to encapsulate it in there because it does not need to be available for the whole class. The only function that should mess with it is `_prepare_host`

src/cript/nodes/primary_nodes/primary_base_node.py
```diff
@@ -1,36 +1,79 @@
...
 from cript.nodes.core import BaseNode
-from cript.nodes.supporting_nodes.user import User
 
 
-class PrimaryBaseNode(BaseNode, ABC):
+def get_uuid_from_uid(uid):
+    return str(uuid.UUID(uid[2:]))
+
+
+class UUIDBaseNode(BaseNode):
```
nh916, line 12: I think if other nodes will inherit from this it might be a better idea to put it somewhere outside of `primary_base_node` because this implies that this base is only for nodes found within the `primary_nodes` directory. It doesn't make sense for any subobjects to inherit from a primary base node

tests/nodes/subobjects/test_software.py
```diff
@@ -21,3 +22,18 @@ def test_setter_getter(complex_software_node):
     assert s2.version == "v0.3.0"
     s2.source = "https://github.com/SSAGESLabs/PySAGES"
     assert s2.source == "https://github.com/SSAGESLabs/PySAGES"
+
+
+def test_uuid(complex_software_node):
+    s = complex_software_node
+
+    # Deep copies should not share uuid (or uids) or urls
+    s2 = copy.deepcopy(complex_software_node)
```
nh916, line 31: can we use the regular `complex_software_node` here without `deepcopy` ?

nh916:
I think it might be a good idea to put `reference` and `software` nodes within the primary nodes directory since they are inheriting from primary nodes file and data model has them as primary nodes

tests/nodes/subobjects/test_software.py
InnocentBug: No, if you use the `complex_software_node` directly, they have the same uuid.
They are even the same python instance.
So there is no point in checking if their uuid id changed as expected.
The deepcopy is absolutely critical to this here.

src/cript/api/api.py
InnocentBug: Yes, agreed if that was true.
But for the construction of node URLs we also need the API_VERSION_STRING.
See primary_base_node:44.

Ideally this would be supplied by the connection to the API at the hand shake, but until then, I wanted it in a central location, where it can be accessed from everywhere we need it.

src/cript/nodes/primary_nodes/primary_base_node.py
InnocentBug: done

InnocentBug:
> @InnocentBug I have a bunch of files that are not merged yet and I am afraid this would create a lot of merge conflicts and we would lose some code when trying to resolve the merges.
> 
> the only thing that I can think of is waiting until I get all the tests to pass and then try to merge these branches into each other, what are your thoughts?

This is based off of `validate-nodes` and targeted for `validate-nodes`.
Plus I just merged `validate-nodes` into it (plus ./trunk fmt).
So there shouldn't be any conflicts.

src/cript/api/api.py
nh916: oh I see, it seems like we are reconstructing the host while we are making the UUID, I think a better idea would be if we have the host cached somewhere and after the API is set every place gets it directly from the host and doesn't have to change anything because if tomorrow the host changes from `host/api/v1` to `host/my_api/v2` then we would just change the way the host is constructed in one file and all other files that are using it would fall into place correctly

src/cript/nodes/primary_nodes/primary_base_node.py
nh916: I honestly, agree, having one class per file makes it easier to read

tests/nodes/subobjects/test_software.py
nh916: okay got it!

nh916:
> > @InnocentBug I have a bunch of files that are not merged yet and I am afraid this would create a lot of merge conflicts and we would lose some code when trying to resolve the merges.
> > the only thing that I can think of is waiting until I get all the tests to pass and then try to merge these branches into each other, what are your thoughts?
> 
> This is based off of `validate-nodes` and targeted for `validate-nodes`. Plus I just merged `validate-nodes` into it (plus ./trunk fmt). So there shouldn't be any conflicts.

yeah I saw that it didn't have any merge conflicts and it was a huge relief. way to go!

src/cript/api/api.py
InnocentBug: That is possible too.
It would be a api function somehow.
And hopefully in the future the API comminucates this on hand shake.
Do you want to open a low priority trello ticket for this?

In the mean time, I can implement this as an API member function, to get host/api/version from the cached API.

nh916: I'm thinking since we have a cached API the other functions can just get the host from the api property?
what do you think? does that work/

nh916: I can add this in pretty quickly right now as well

InnocentBug: Just pushed that

src/cript/api/api.py
```diff
@@ -38,52 +38,20 @@ def _get_global_cached_api():
...
-
-
 class API:
     """
     ## Definition
     API Client class to communicate with the CRIPT API
     """
 
-    _host: str = ""
+    _raw_host: str = ""
```
nh916, line 47: how come we are keeping this?
do we need this anywhere else?

src/cript/api/api.py
```diff
@@ -38,52 +38,20 @@ def _get_global_cached_api():
...
+    _https_warning_sent: bool = False
     _token: str = ""
     _vocabulary: dict = {}
     _db_schema: dict = {}
     _http_headers: dict = {}
+    _api_handle: str = "api"
+    _api_version: str = "v1"
```
nh916, line 54: I think this is okay but I don't see it necessary to separate out `_api_handle` and `_api_version` for now, but I think this is also fine

nh916, line 54: is there anyway that we can encapsulate everything needed for the api host into one or two function, so nothing else needs to have access to it because I don't think anything else needs it

src/cript/api/api.py
```diff
@@ -194,6 +162,14 @@ def disconnect(self):
         global _global_cached_api
         _global_cached_api = self._previous_global_cached_api
 
+    @property
+    def api_handle(self):
+        return self._api_handle
+
+    @property
+    def api_version(self):
+        return self._api_version
```
nh916, line 171: I don't think there is a scenario that either of these two will be needed and we'll always use only a single host

changing a single small string of `/api/v1/` is pretty simple that I wouldn't put any variables for it, although we could

src/cript/api/api.py
```diff
@@ -217,7 +193,21 @@ def host(self):
         https://criptapp.org/api/v1
         ```
         """
-        return self._host
+        host = self._raw_host.strip()
```
nh916, line 196: how come this was removed from `_perpare_host` function and put here?

src/cript/api/api.py
```diff
@@ -38,52 +38,20 @@ def _get_global_cached_api():
...
-
 class API:
     """
     ## Definition
     API Client class to communicate with the CRIPT API
     """
 
-    _host: str = ""
+    _raw_host: str = ""
+    _https_warning_sent: bool = False
```
nh916, line 48: why do we need to know if we gave a warning already or not?

I think if the user sees the warning multiple times it is okay, there are other libraries where they give a warning multiple times

I think if we had it like before with prepare host a lot of these could just be encapsulated into there safely and neatly in one place and warning would only get triggered once because it would only hit prepare host on `__init__`

src/cript/nodes/uuid_base.py
```diff
@@ -0,0 +1,49 @@
...
+    @property
+    def uuid(self) -> uuid.UUID:
+        return uuid.UUID(self._json_attrs.uuid)
+
+    @property
+    def url(self):
+        from cript.api.api import _get_global_cached_api
+
+        api = _get_global_cached_api()
+        return f"{api.host}/{self.uuid}"
```
nh916, line 44: this is perfect!

src/cript/api/api.py
InnocentBug: we are keeping it now, because it is so trivial to construct the host from it.
It might be helpful to have for debugging at some point. (You actually know what the user submitted, and something goes wrong with constructing the url from it.)

src/cript/api/api.py
InnocentBug: I did this to be future prove.
At some point these two can be send from the back end to us at handshake. And I quite sure it would be two separate things then. It also is pretty clear.

InnocentBug: This is encapsulated.
These are private memebers of the class. That is about as encapsulated as you can get.

src/cript/api/api.py
InnocentBug: I wanted them as property, because one day we might need to do some magic to get it from back end instead of just reading a string.
And this way we have it read-only.

src/cript/api/api.py
InnocentBug: And extra function for this seems overkill.
I get, why you did it, since you only need to construct it once.
But I liked it being more dynamic.

src/cript/api/api.py
InnocentBug: We can get rid of it.
I just thought issueing many warning might be annoying.

src/cript/api/api.py
nh916: hmm yeah I think that could work

src/cript/api/api.py
InnocentBug: Also, since we are now using member attributes of the API class for the host construction.
I either had to pass an API object to the `_prepare_host` function, or the prefix separately.
And passing an API object into the function, is basically the same as a member function.
So, I made it a member function right away.
And to no clunk up the class namespace anymore then necessary, the host property seems a good place for it.

src/cript/api/api.py
nh916: we could encapsulate them within a function and not have the rest of the class see it or need it, we only set the host variable for the class and the rest of the class never sees this

src/cript/api/api.py
nh916: this is pretty small change and when we need it in the future lets add it then because we might just add this and never need this or worst it doesn't help in the future at all

src/cript/api/api.py
nh916: but it's still only constructed once, how come host needs to be dynamic?

src/cript/api/api.py
InnocentBug: k

src/cript/api/api.py
InnocentBug: no, now it is constructed every time you call `API.host`

src/cript/api/api.py
nh916: this makes it confusing for anyone reading the code because there is 2 different host and it's not needed, debugging the string is going to be pretty easy, I can't think of a scenario when this is needed

InnocentBug: No, there is only one host.
Especially user facing there is only one host.

I don't know you asked me to figure out a better way of handling this.
This is my solution. Hard coding the api modifier was a bad solution, so we needed something better.

src/cript/api/api.py
nh916: how come we don't want to construct it once and save it in the class as a class variable?

nh916: I strongly recommend having everything for the host within `_prepare_host` where it was encapsulated nice and neatly

InnocentBug: That is a possible solution too.
But it needs one more function to debug.
My feeling was that the namespace was pretty full already, so I got rid of one function.

InnocentBug: > I strongly recommend having everything for the host within `_prepare_host` where it was encapsulated nice and neatly.

I disagree here.
You encapsulated it too much.
When I needed access to the API version string and where it was constructed, it took me a while to find it.

InnocentBug: I gave you a solution that just put the hard coded string into a global variable.
Like anything hard coded should be.
But you didn't like that approach either.

nh916 (approved):

PR #92, opened 2023-05-09
host idea

nh916:
# Description

## Checklist


InnocentBug (approved):
It's fine. Not the biggest fan, but I am ok with it.
Passing all these arguments is a bit clunky but not prohibitive.
And we can live with the risk that `API.host` and `API.api_handle` /`API.api_version` differ.
Not great, but also not worth fighting about.

InnocentBug (approved):

PR #95, opened 2023-05-11
Feature: `API.upload_file()` & `API.download_file()`

nh916:
# Description
AWS S3 upload file and download file working correctly

## Changes
* brought all AWS S3 variables to class level
* file upload and download working
    * during file upload the file name is changed to `file_name_uuid4_hex.extension`
        * using UUID4 hex to not have dashes in the file name and have it be a bit cleaner
* wrote a test for both file upload and download
* changing the `_BUCKET_DIRECTORY_NAME` of `cript.API` within `conftest.py` to be sure tests only go into `tests/` directory of the AWS S3 storage

## Tests
* tests pass locally but since it needs a token it will not pass on GitHub actions
* testing file upload to our AWS S3 bucket is problematic because  every time we run this test we are uploading a new file to AWS S3, this could increase our storage costs
    * I have already ran it a bunch of times during testing and development

## Known Issues
* the file URLs are available forever after they have been uploaded
    * this could be a security risk, but the I am not fully sure if it is or how
    * we could add a time limit to the URL, but I think that would break our software because later users won't be able to get their files via that same URL
       * code to set a time limit on the uploaded file URL:
        ```python
         url_expire_time = datetime.timedelta(minutes=5)  # URL will expire in 5 minutes
        s3_file_url = self._s3_client.generate_presigned_url(
            'get_object',
            Params={'Bucket': self._BUCKET_NAME, 'Key': object_name},
            ExpiresIn=url_expire_time.total_seconds()
        )
        ```
* `test_upload_and_download_file` is a bit complex and hard to understand, but I have docstrings and comments
* not sure how good the UX is for file upload and download yet
    * the UX feels a bit weird to me because the user would have to pass in the directory, file name, and extension to the function which the user might be less intuitive or the user might not know the file extension
* further error handling for both upload and download might be needed
* The biggest issue with this that I see is that the AWS S3 bucket sends back bytes and does not tell me the file extension.
    * We could try to find the file extension from the file node `extension` attribute and add it onto there, but not sure about it currently

## Notes
* I was thinking of encapsulating the `_s3_client` client in the `upload_file` method because it is only used there, but all other variables for s3 are already at class level and I get the feeling that if we want to do anything else with the s3 client then we have to move it to the class level again. Switching from class level to method would be very easy and I can do it if we think that is a better way to go.

| WARNING: After the PR is merged it will require a token from the frontend and all tokens are verified by AWS and invalid tokens will break API tests |
| --- |

### Thoughts
[I think file upload and download should be at API level](https://trello.com/c/SHsl4plZ)

I think a workflow like this might be easier:

1. We could create an API endpoint like `https://mycriptapp.org/files/upload`.
2. Then, we could send an HTTP POST request to that URL with the file.
3. The API would store the file wherever it wants and respond back with the URL of the stored file.
4. The file URL from the API can then be used as a file node source attribute to generate a giant JSON with only web file sources and no local file sources anymore.

This approach offers several advantages:

- The API has control over the files, ensuring consistent conventions.
- It promotes uniformity, so the frontend doesn't save files like `_file_123456_789012.pdf`, and the SDK doesn't save it like `123456_file_name_789012_.pdf`.
- It allows for easy switching of storage providers (e.g., from AWS to GCP) without impacting clients interacting with the API.

I think this might be a good idea as it simplifies the workflow, promotes consistency, and allows for flexibility in the future.


nh916:
@InnocentBug let me get your thoughts on this when you get a chacne

src/cript/api/api.py
```diff
@@ -52,6 +57,14 @@ class API:
     _api_handle: str = "api"
     _api_version: str = "v1"
 
+    # AWS S3 constants
+    _REGION_NAME: str = "us-east-1"
```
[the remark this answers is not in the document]
InnocentBug: Ideally this info is communicated to us during the initial hand shake with back end.

src/cript/api/api.py
```diff
@@ -429,6 +470,126 @@ def save(self, project: Project) -> None:
         if response["code"] != 200:
             raise CRIPTAPISaveError(api_host_domain=self._host, http_code=response["code"], api_response=response["error"])
 
+    def upload_file(self, file_path: Union[Path, str]) -> str:
```
InnocentBug, line 473: not sure this should be a public function.
Typically users should see or use this directly.

src/cript/api/api.py
```diff
@@ -429,6 +470,126 @@ def save(self, project: Project) -> None:
...
+
+        # Generate a presigned URL to access the file from S3 that is available forever
+        s3_file_url = self._s3_client.generate_presigned_url(
+            "get_object",
+            Params={"Bucket": self._BUCKET_NAME, "Key": object_name},
+        )
+
+        return s3_file_url
+
+    def download_file(self, file_url: str, destination_path: Union[Path, str]) -> None:
```
InnocentBug, line 546: also, should be user facing.

What should be user facing though is a spot where all files are downloaded to.
I would say, we initially have a default either `.` or `~/Downloads` but users can set with the API class or an env variable where all files should be stored that are downloaded.

InnocentBug (changes requested):
This is a very manual approach to file up and download.

I have explained before, that the `File` node class should handle this automatically for the user.
Otherwise we are running the risk, that users forget to upload their files, or upload wrong files etc.

It is relatively straight forward to have all of this handle automatically.
So we should do that. (Even the old SDK did that.)

Also, the API class is getting out of control.
In version 2, it needs refactoring. 
It should hold a number of sub class elements that implement and do most of the work.
Like a DB schema class, a file up/down load class. etc.

src/cript/api/api.py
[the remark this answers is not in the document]
nh916: for me ideally the API would handle the file upload and download and we would just make a request to it for the reasons stated in the PR description.

From what I can tell, I am not sure if there will be an initial hand shake with the API

nh916:
> This is a very manual approach to file up and download.
> 
> I have explained before, that the `File` node class should handle this automatically for the user. Otherwise we are running the risk, that users forget to upload their files, or upload wrong files etc.
> 
> It is relatively straight forward to have all of this handle automatically. So we should do that. (Even the old SDK did that.)
> 
> Also, the API class is getting out of control. In version 2, it needs refactoring. It should hold a number of sub class elements that implement and do most of the work. Like a DB schema class, a file up/down load class. etc.

---

1. I agree and I am planning on doing that for the `File node` this is only the initial PR so that we have the API client side done, and then we can use the API client methods within the `File node` for upload and download

2. agreed, for V2 we need a to start looking into refactoring strategies as API client class is getting big

src/cript/api/api.py
nh916: I was not fully sure either, but I thought that it made sense to me to have `upload_file` to be public so that if the user wants to upload it themselves within their script they can

and within the file node we check that if the file source starts with `http` then we accept it, otherwise it would be a local file path and we upload that to our AWS S3, get a link, and put the link into the file source

src/cript/api/api.py
nh916: I was not sure about `~/Downloads` because I wasn't sure if that would work with all OS

I can put it within the cwd

nh916:
SDK needs a token from API that has access to AWS S3, otherwise every API test will fail because during initialization when it tries to create an S3 client a fake token will throw an error.

This is blocked until we can get this resolved so we can continue working on the tests to pass within CI

nh916:
This test needs a valid API token to work correctly with the cloud storage provider

nh916:
too many conflicts, making a new PR: https://github.com/C-Accel-CRIPT/Python-SDK/pull/156

PR #96, opened 2023-05-11
get material identifiers dynamically

nh916:
# Description
wrote `cript.API.get_vocab_by_category()` method that can be used by both users and developers working on the CRIPT Python SDK

* Users can use it to see all available vocabulary for a given category right within their IDE without ever having to leave their code and go to the website
* CRIPT Python SDK developers can use it to get the controlled vocabulary by category as they need validate things

## Changes
* wrote the api method for `get_vocabulary_by_category`
* wrote tests for `get_vocabulary_by_category`
* using `get_vocabulary_by_category` within material identifiers
* vocabulary categories have been converted from a list of available key names (str) to enums
    * if the user wants to see the available options for a vocabulary category they can see it right from their IDE with
        
        ```python
        print(api.get_vocab_by_category(category=ControlledVocabularyCategories.FILE_TYPE))
        ```
* removed `InvalidVocabularyCategory` because since vocabulary categories are enums and they get accessed via enums, it is no longer applicable
* added `ControlledVocabularyCategories` enum to package for easy access

## Tests
* `test_deserialize_material_from_json` passing ✅ 
* wrote a rough `test_get_vocabulary_by_category` which is passing ✅ 

## Known Issues
* `get_vocabulary_by_category()` works with a string, and needs to instead work with enums to be easier to use and would result in better UX for users and less errors
* `_get_vocabulary()` method should be later upgraded to call `get_vocabulary_by_category()` and have all logic in `get_vocabulary_by_category()` and it just calls it a bunch of times with all the possible vocabulary found in the ValidVocabulary enum
* `ControlledVocabularyCategories` name is a bit long and could maybe be shortened
    * shortening the name now or later should be a breeze

## Notes
* the enum class name `ControlledVocabularyCategories` might be too long


nh916:
@InnocentBug I wrote up a rough idea of an upgrade for material identifiers, let me know what you think and if you like it, I can clean it up a bit, improve it, and make it into a proper PR.

I think this part I've written in a crummy way, that can probably use some clean up

```python
    def _from_json(cls, json_dict: dict):

        from cript.api.api import _get_global_cached_api
        api = _get_global_cached_api()

        # get JSON vocabulary of all "material_identifier_key"
        all_identifiers: List[dict] = api.get_vocab_by_category("material_identifier_key")

        # placeholder for what eventually will be all the names of material identifiers
        # e.g. "smiles", "bigsmiles", etc.
        all_identifiers_list: List[str] = []
```

src/cript/api/api.py
```diff
@@ -258,6 +259,33 @@ def _get_vocab(self) -> dict:
 
         return self._vocabulary
 
+    # TODO needs to work with enums instead of strings
+    def get_vocab_by_category(self, category: str) -> List[dict]:
```
InnocentBug, line 263: I think it would be best to have this not contact the API all the time.
This is probably going to be called a lot if  we use if for material deserialization and we could like over load the back end.

We have a mechanism to cache the vocab already, it should use that instead.
(Even for a first version, let's not do the work multiple times.)

src/cript/nodes/primary_nodes/material.py
```diff
@@ -445,27 +445,24 @@ def property(self, new_property_list: List[Any]) -> None:
...
-            "vendor",
-        ]
+
+        from cript.api.api import _get_global_cached_api
+        api = _get_global_cached_api()
+
+        # get JSON vocabulary of all "material_identifier_key"
+        all_identifiers: List[dict] = api.get_vocab_by_category("material_identifier_key")
+
+        # placeholder for what eventually will be all the names of material identifiers
```
InnocentBug, line 455: The logic here is fine.
But the comment really confused me.
The variable is not really a placeholder.
It actually is the list of all string names of identifiers.
You fill them right away with a for loop.

If you need a comment to explain that, consider using list compaction and no comment.

src/cript/nodes/primary_nodes/material.py
```diff
@@ -445,27 +445,24 @@ def property(self, new_property_list: List[Any]) -> None:
...
+
+        # loop through and take all "name" from every dict and append it to `all_identifiers_list`
+        for identifier in all_identifiers:
+            all_identifiers_list.append(identifier.get("name"))
 
         identifier_argument = []
-        # Convert identifiers to to a list
-        for identifier in all_identifier_list:
+        # Convert identifiers to a list
+        for identifier in all_identifiers_list:
```
InnocentBug, line 465: Yes, this is great.
Automatically assembling the possible list.
We rely on the vocab to be complete here, but I think as is a fair assumption.
It would be a serious bug of the back end if the material allows an identifier, that is not in the vocab.

tests/api/test_api.py
```diff
@@ -95,6 +95,46 @@ def test_is_node_schema_valid(cript_api: cript.API) -> None:
...
+
+    # test response is a list of dicts
+    assert isinstance(material_identifier_vocab_list, list)
+
+    # test it has basic material identifiers
+    has_alternative_names = False
+    has_bigsmiles = False
+    has_smiles = False
+
+    for identifier in material_identifier_vocab_list:
```
InnocentBug, line 122: There should be a better way to test this.

something like `assert "bigsmiles" in material_identifier_list`.

InnocentBug (changes requested):
This is a great idea.
And almost there.

It could be a little more clear at some points, but that should be simple to upgrade.

And we need need to use the cached vocab instead of contacting the back end all the time.

src/cript/api/api.py
nh916: You are 100% right, great catch, let me add that in

tests/api/test_api.py
nh916: that is way way cleaner!

I'll add it in

nh916: actually, I don't think I can do that because I it is a list of dicts

```json
[
   {
      "description":"preferred name",
      "name":"preferred_name",
      "validator":"None",
      "value_type":"string"
   },
   {
      "description":"additional names",
      "name":"alternative_names",
      "validator":"None",
      "value_type":"list[string]"
   },
   {
      "description":"Chemical ID",
      "name":"chemical_id",
      "validator":"None",
      "value_type":"string"
   },
   {
      "description":"bigSMILES line notation",
      "name":"bigsmiles",
      "validator":"None",
      "value_type":"string"
   },
   {
      "description":"simplified molecular-input line-entry system (SMILES)",
      "name":"smiles",
      "validator":"None",
      "value_type":"string"
   },
   {
      "description":"chemical formula",
      "name":"chem_formula",
      "validator":"None",
      "value_type":"string"
   },
   {
      "description":"chemical formula of repeating unit",
      "name":"chem_repeat",
      "validator":"None",
      "value_type":"list[string]"
   },
   {
      "description":"PubChem CID",
      "name":"pubchem_cid",
      "validator":"None",
      "value_type":"integer"
   },
   {
      "description":"IUPAC International Chemical Identifier",
      "name":"inchi",
      "validator":"None",
      "value_type":"string"
   },
   {
      "description":"a hashed version of the full InChI",
      "name":"inchi_key",
      "validator":"None",
      "value_type":"string"
   },
   {
      "description":"amino acid sequence",
      "name":"amino_acid",
      "validator":"None",
      "value_type":"string"
   },
   {
      "description":"company the material was obtain from",
      "name":"vendor",
      "validator":"None",
      "value_type":"string"
   },
   {
      "description":"lot or batch number",
      "name":"lot_number",
      "validator":"None",
      "value_type":"string"
   }
]
```

src/cript/nodes/primary_nodes/material.py
nh916: yeah I think that is mostly okay, if something is wrong with the backend that is out of our hands

nh916:
@InnocentBug I think enums are a good idea for the controlled vocabulary because then there cannot be any mistakes in spelling, but not sure if it is the best choice or not

src/cript/api/api.py
nh916: done! ✅

src/cript/nodes/primary_nodes/material.py
nh916: yeah I agree, it's a bit confusing and a bit messy so I honestly don't love it, I wasn't sure how to clean it up.

I could do something like this, but I feel like this is even more confusing

```python
        all_identifiers_list = [
            identifier.get("name")
            for identifier in api.get_vocab_by_category(cript.ControlledVocabularyCategories.MATERIAL_IDENTIFIER_KEY)
        ]
```

nh916:
@InnocentBug I converted the list of string vocabulary categories to enums, which I think could make it easier to use and reduce errors, what are your thoughts?

nh916:
this entire code just feels a bit messy and not sure how to clean it up either

```python
    @classmethod
    def _from_json(cls, json_dict: dict):

        from cript.api.api import _get_global_cached_api
        api = _get_global_cached_api()

        all_identifiers_list = [
            identifier.get("name")
            for identifier in api.get_vocab_by_category(cript.ControlledVocabularyCategories.MATERIAL_IDENTIFIER_KEY)
        ]

        # pop "name" from identifiers list because the node has to have a name
        all_identifiers_list.remove("name")

        identifier_argument = []
        # Convert identifiers to a list
        for identifier in all_identifiers_list:
            if identifier in json_dict:
                identifier_argument.append({identifier: json_dict[identifier]})
                del json_dict[identifier]
            json_dict["identifiers"] = identifier_argument

        return super()._from_json(json_dict)
```

I think some docstrings could help though

nh916:
@InnocentBug what are your thoughts on these docstrings, am I getting it correctly or is it off?

```python

    def _from_json(cls, json_dict: dict):
        """
        Create a new instance of a node from a JSON representation.
        Parameters
        ----------
        json_dict : dict
            A JSON dictionary representing a node
        Returns
        -------
        node
            A new instance of a node.
        Notes
        -----
        required fields in JSON:
        * `name`: The name of the node
        optional fields in JSON:
        * `identifiers`: A list of material identifiers.
            * If the `identifiers` property is not present in the JSON dictionary,
            it will be set to an empty list.
        """
```

tests/api/test_api.py
InnocentBug: do a list comprehension exactly you do it before in the actual code.
Then you can do simple `in` testing.

nh916: agreed, that is cleaner

done! ✅

InnocentBug (approved):

src/cript/api/vocabulary_categories.py
```diff
@@ -1,29 +1,91 @@
...
+
+class ControlledVocabularyCategories(Enum):
+    """
+    All available CRIPT controlled vocabulary categories
+
+    Controlled vocabulary categories are used to classify data.
+
+    Attributes
+    ----------
+    ALGORITHM_KEY: str
```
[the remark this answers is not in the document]
nh916: that is a great point, I'll look into it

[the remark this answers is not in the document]
nh916: I agree, this is a lot to type out and keep updated.

I've researched this a bit and I can't find any obvious or good solutions. I put the question on a few different forums and I am waiting for their answers to see if there is something that I've missed, but I am pessimistic on that. 

I think it is possible to get all enum attributes with sphinx, but I have not found anything for mkdocs.

here's what I am thinking:
I'm guessing the controlled vocabulary categories probably won't have to change that often for now so we wont have to update the documentation any time soon and this one is already written

if the controlled vocabulary categories do start changing often then we'll need to get the categories dynamically from the API and this enum class would become moot and we'll need a different solution.

what are your thoughts?

> I've put in a [request for querying of vocab categories](https://trello.com/c/EGS145NJ), but it is currently not being supported

nh916:
@InnocentBug you think I should change the enum class name to something else `ControlledVocabularyCategories` because it feels a bit long to me but it is explicit

src/cript/api/vocabulary_categories.py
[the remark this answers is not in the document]
InnocentBug: That makes sense.
It is a bummer, but nothing we can do right now.
So let's push that on the back log.

InnocentBug:
> @InnocentBug you think I should change the enum class name to something else `ControlledVocabularyCategories` because it feels a bit long to me but it is explicit

I like explicit over brevity.
That is why we have IDEs with autocomplete.

src/cript/api/vocabulary_categories.py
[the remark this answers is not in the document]
nh916: agreed!

hopefully that can be upgraded later

nh916:
> > this entire code just feels a bit messy and not sure how to clean it up either. or is it okay?
> > ```python
> >     @classmethod
> >     def _from_json(cls, json_dict: dict):
> > 
> >         from cript.api.api import _get_global_cached_api
> >         api = _get_global_cached_api()
> > 
> >         # get material identifiers keys from API
> >         # create a list with all material identifiers in the "name" column
> >         # eg ["smiles", "bigsmiles", etc.]
> >         all_identifiers_list = [
> >             identifier.get("name")
> >             for identifier in api.get_vocab_by_category(cript.ControlledVocabularyCategories.MATERIAL_IDENTIFIER_KEY)
> >         ]
> > 
> >         # pop "name" from identifiers list because the node has to have a name
> >         all_identifiers_list.remove("name")
> > 
> >         identifier_argument = []
> >         # Convert identifiers to a list
> >         for identifier in all_identifiers_list:
> >             if identifier in json_dict:
> >                 identifier_argument.append({identifier: json_dict[identifier]})
> >                 del json_dict[identifier]
> >             json_dict["identifiers"] = identifier_argument
> > 
> >         return super()._from_json(json_dict)
> > ```
> 
> I agree, it became a little messy over time. We should probably outsource the Identifier conversion into a separate function, maybe even module.
> 
> Would be good to have serialization and deserialation functions for this in one place. But this doesn't have priority for me.


---
---

That is a great idea, I put it inside of `src/cript/nodes/utils/material_identifier_deserialization.py` let me know your thoughts

InnocentBug (approved):

PR #98, opened 2023-05-16
Read `host` and `token` from config.json

nh916:
# Description
created feature to be able to read `host` and `token` from `config.json`

Works with both
absolute paths 
```bash
C:\\users\username\onedrive\desktop\config.json
```

Path object (recommended)
```python
(Path(__file__) / Path("../../config.json)).resolve()
```

## Changes
* added a few lines of code to constructor of api.py
* added `config.json` to `.gitignore`

## Tests
* wrote a simple test for it

## Screenshot
Updated documentation
image


nh916:
test that is failing is because it expects the db schema to have `70` sections but it has `74` currently

src/cript/api/api.py
```diff
@@ -85,18 +86,44 @@ class API:
...
+            "token": "I am token"
+        }
+        ```
+
+        `my_script.py`
+        ```python
+        from pathlib import Path
+
+        # create a file path object of where the config file is
+        config_file_path = Path(__file__) / Path('./config.json')
```
InnocentBug, line 121: This seems overly complicated.
I assume this to be compatible with windows?

I would recomment `os.path.join` and `os.path` for this.

src/cript/api/api.py
```diff
@@ -111,6 +138,8 @@ def __init__(self, host: Union[str, None], token: [str, None]):
             You can find your personal token on the cript website at User > Security Settings.
             The user icon is in the top right.
             If `None` is specified, the token is inferred from the environment variable `CRIPT_TOKEN`.
+        config_file_path: Union[Path, str]
```
InnocentBug, line 141: We should clearly document which of the option takes precedent if multiple are specified.

src/cript/api/api.py
```diff
@@ -135,6 +164,17 @@ def __init__(self, host: Union[str, None], token: [str, None]):
             Instantiate a new CRIPT API object
         """
 
+        if config_file_path and not host and not token:
```
InnocentBug, line 167: Maybe this should be a little more elaborate.
I.e. if a user specifies host via an argument (or env variable) but the `token` via the config file, this should be OK.

If you are asked me, config file should take 3. prioty.

1. direct arguments.
2. if direct arguments, try env variable
3. if direct argument and env variable not available take config file.

InnocentBug (changes requested):

src/cript/api/api.py
```diff
@@ -85,18 +86,44 @@ class API:
     _db_schema: dict = {}
     _http_headers: dict = {}
 
-    def __init__(self, host: Union[str, None], token: [str, None]):
+    def __init__(self, host: Union[str, None] = "", token: [str, None] = "", config_file_path: Union[Path, str] = ""):
```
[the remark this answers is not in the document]
nh916: I was debating between `None` or `""` and I thought if I put it as `None` then the user can do `cript.API()` without specifying anything, but that could also work as well because then the user can create a CRIPT API easier.

I can change it to `None` for now, and we can change it later if needed

src/cript/api/api.py
nh916: I think that list makes sense. Mine was config file as priority and I can swing either way for direct and env vars. I'm thinking because config file is easy to setup and encourages user to take out arguments out of their code so they can easily share it.

what are your thoughts, I can swing this either way, I don't have a strong preference one way or another

src/cript/api/api.py
nh916: I think this is the recommended way because that way the file path would be found regardless of the operating system, and if I am not mistaking I think this it the recommended way of getting a file path.

From when I was looking into it, I think I found that `pathlib.Path` is built on top of `os.path` and is often considered a nicer interface.

I wasn't sure if I should just accept anything and just convert it under the hood to a `Path` object or not.  The user has a bunch of options they can choose from. 

I can also put all the possibilities in the documentation as well if we want

---

## Usage
### absolute path string
```python
config_file="C:\\Users\username\desktop\crip\my_file"
```

### relative path
```python
config_file_path ="./.config.json"
```

### pathlib Path object
```python
# always gets the correct file regardless of which directory the code is ran from
config_file_path = (Path(__file__) / Path('./config.json')).resolve
```

### os.path.join
```python
# always gets the correct file regardless of which directory the code is ran from
file_path = os.path.join(os.path.dirname(__file__), "config.json")
```

### os.path
```python
# always gets the correct file regardless of which directory the code is ran from
file_path = os.path.abspath(os.path.join(os.path.dirname(__file__), 'config.json'))
```

---

I think they would all have the similar levels of difficulty

## Difficulty
### `os.path.join`
```python
import os

config_path = os.path.join(os.path.dirname(__file__), 'config.json')
```

### `os.path`
```python
import os

config_path = os.path.abspath(os.path.join(os.path.dirname(__file__), 'config.json'))
```

### `pathlib.Path`
```Python
Path(__file__) / Path('./config.json')
```

src/cript/api/api.py
nh916: sure, as soon as we get that figured out I can add it in

nh916:
I think I can put the host and token initialization from the constructor into a utility function and clean up our constructor a bit

src/cript/api/api.py
nh916: I think for v1, I'm not going to allow for reading the host from config and token from environment variable, either put all into environment variable, config, or directly and in next versions we can start allowing for more flexbility

src/cript/api/api.py
nh916: I think the way I wrote it with utils folder makes it very flexible so we can change up the order whenever we want

src/cript/api/api.py
InnocentBug: ok, I was just confused, because I saw you using `/` which is OS specific.

InnocentBug:
I can approve when tests are passing.

src/cript/api/utils/_get_host_and_token.py
```diff
@@ -0,0 +1,58 @@
+import json
```
InnocentBug, line 1: I saw that you used `utils/*.py` in the other branch about the identifiers too.

Can we please consolidate this with `util.py`?
Having `util` and `utils` as separate things is confusing.


I resolved this in the validate node branch already like this: b697be6e4315ba31d8

nh916: do you want to have a utils directory with separate utils files or are you thinking a single utils.py file for everything?

nh916: do you want to have a utils directory with separate utils files or are you thinking a single utils.py file for everything?

InnocentBug: What I did is, I renamed the `utils` directory into `util`  and the `util.py` file into `util/__init__.py`.
Making the existing `util.py` the core of the new `util` module. (Hence imports of the old stuff doesn't change.)

And from there we can add more files into the module.
Seems like a clean approach to me.

nh916: I see that you have utils in `src/cript/node/utils/` are you thinking of putting the api utils in that file?

InnocentBug: No, I was thinking that this module is strictly only for nodes.

But we can mimic that structure for the API as well.

nh916: ooooh okay gotcha! okay I'll do that for api util folder too!

nh916:
@InnocentBug I think I'm going to make a new PR because I was indecisive if I was putting this PR into `validate` branch or `develop` branch, so this PR history became a mess, but luckily it is a small change, so I can add the changes to a new PR and put it into develop branch

PR #99, opened 2023-05-16
PR template updated

nh916:
# Description
added a check list item to remind anyone updating the code to also update the documentation that pertains to their code


nh916:
test that is failing is because it expects the db schema to have `70` sections but it has `74` currently

InnocentBug (approved):

PR #102, opened 2023-05-18
Feature: API config.json

nh916:
# Description
created feature to be able to read `host` and `token` from `config.json`

Works with both
absolute paths 
```bash
C:\\users\username\onedrive\desktop\config.json
```

Path object (recommended)
```python
(Path(__file__) / Path("../../config.json)).resolve()
```

## Changes
* added a few lines of code to constructor of api.py
* added `config.json` to `.gitignore`

## Tests
* wrote a simple test for it

## Screenshot
Updated documentation
image

## Notes
I've written this PR a few times because I am not sure which branch to put it into. It makes sense to go into develop branch because it is a feature, but if I put it there it might create issues later or have merge conflicts because there are already changes to the API class in the validate branch. I think I am just going to put it into the validate branch and continue.


nh916:
this is a PR that is similar to the [previous PR for this feature](https://github.com/C-Accel-CRIPT/Python-SDK/pull/98) but with a clean history that will not mess anything up

nh916:
we have to enable checks on all PR and not just the ones to `develop` or `main` because that would help see if the tests are passing or not

InnocentBug (approved):

PR #103, opened 2023-05-19
updated tests and trunk workflow

nh916:
# Description
updating workflow so that it runs on all PR, so we can see if something is failing if we are merging one unprotected branch to another

## Changes
* changed tests workflow
* trunk workflow

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,6 +11,7 @@ on:
       - main
       - develop
       - trunk-merge/**
+      - "*"
 
 jobs:
   install:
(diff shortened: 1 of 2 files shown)


nh916:
this will need to be also merged into [validate-nodes branch](https://github.com/C-Accel-CRIPT/Python-SDK/pull/78) so that when we are merging branches into it we see the checks.

I don't want to do `git merge` yet because I assume there will be a lot of merge conflicts that we may want to resolve later, but we can put another PR to put this code in there as well

nh916:
the test that is failing is the `db schema test` that is is expecting 70 sections, but is getting 74

PR #108, opened 2023-05-24
Cspell

InnocentBug:
# Description
Add cspell as a checker for trunk.

This should alert us for any misspelled word.
## Changes

```
./trunk check enable cspell
```

## Tests
automatically included in trunk check
## Notes
I haven't actually started to fix all the issues, so for now it fails.
We need to go through the code and fix all misspelling, or teach cspell the missing words.
## Checklist

- [x] My name is on the list of contributors (`CONTRIBUTORS.md`) in the pull request source branch.

PR #109, opened 2023-05-24
added words to dictionary

nh916:
# Description
added some of the common words that is used in Polymer research and CRIPT into the `cspell` dictionary. Left out `subobject` because in majority of places I have misspelled it and it should actually be `sub-object`

## Checklist


InnocentBug:
already handled with #108

PR #112, opened 2023-05-25
Update README.md

nh916:
# Description
updated `README` with `Python SDK documentation` and notes for developers to read the `wiki` and communicate with us via `issues` and `discussions`

## Changes
* added documentation link to it
* added wiki link to it

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ pip install cript
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


nh916:
the one test that is failing is because it expects db schema to have 70 sections, but it has 74

InnocentBug (approved):
we should remove the length of DB schema test

nh916:
> we should remove the length of DB schema test

I fully agree! I removed it in the `validate` branch I think, but that has not been merged yet

PR #118, opened 2023-05-25
Slimming tests

nh916:
# Description
Having a lot of tests can be good, but having a lot will create maintenance harder, more timely, and overall more costly. This PR addresses that issue by removing tests that are unneeded, already being captured somewhere else, and overall duplicates. I think these tests can be removed without much disadvantage.

## Changed Test Files
* `Computation`
* `Computation Process`
* `Data`
* `Experiment`
* `Material`
* `Process`
* `Reference`

## Checklist


InnocentBug:
I don't know we don't need to close it yet.
Maybe we think about it a bit more?

nh916:
> I don't know we don't need to close it yet. Maybe we think about it a bit more?

okie dokes, I am okay with that!

nh916:
@InnocentBug I'm thinking of removing this PR  because I think we both like to keep both minimal and maximal tests

InnocentBug:
ok, let's close it.

The merge hopefully takes some of this pain away.

PR #122, opened 2023-05-26
Add beartype checking

nh916:
# Description
adding runtime type checker to SDK nodes to help catch errors

## Changes
* added [beartype](https://github.com/beartype/beartype) to `requirements.txt`
     * for full detailed explanation please visit the [Trello ticket](https://trello.com/c/AsPEQPqQ)
 * added beartype type checker to all node constructor and getter and setters
     * putting type checker after `@property` decorator because decorators are executed from top to bottom, so property has be executed first
 * removed `_validate_keyword` and `_validate_identifiers` from `material.py` because identifiers and keyword checking is done via db schema

## Known Issues
* the errors that are often given to the user are not great
    * example of a type error
    ```bash
    File "<@beartype(__main__.add_numbers_beartype) at 0x2576bd16dd0>", line 41, in add_numbers_beartype
    beartype.roar.BeartypeCallHintParamViolation: Function __main__.add_numbers_beartype() parameter b=[2] violates type hint 
    <class 'int'>, as list [2] not instance of int.
    ```

## Notes
* down the line for another upgrade I think it would be a good idea to move from [beartype](https://github.com/beartype/beartype) to [pydantic 2](https://docs.pydantic.dev/dev-v2/) for both type checking and serailization/deserialization

## Checklist


InnocentBug:
let's address the conflicts before review.

InnocentBug:
I will have to address the found issues before we can merge

InnocentBug:
This is include now in #151 ?

PR #125, opened 2023-05-27
added complex material test

nh916:
# Description
I noticed that the `Material` node only has a simple test, so I replaced it with a complex test that tests all of the constructor

## Tests
That material test is still passing as it was before

### DIFF
--- a/tests/nodes/primary_nodes/test_material.py
+++ b/tests/nodes/primary_nodes/test_material.py
@@ -5,19 +5,29 @@ from util import strip_uid_from_dict
 import cript
 
 
-def test_create_simple_material() -> None:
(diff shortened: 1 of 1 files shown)


InnocentBug (approved):

nh916:
I replaced the simple test with a complex test, from the other comment on the other PR we both agreed that a simple test is good to have as well. You think I should add the simple test and have a complex test or just one is fine?

InnocentBug:
> I replaced the simple test with a complex test, from the other comment on the other PR we both agreed that a simple test is good to have as well. You think I should add the simple test and have a complex test or just one is fine?

This should be fine.
It is not that complex here.

PR #131, opened 2023-05-31
created docs_ci_check

nh916:
# Description

This is a bit duplicate, but I think this would be a good check for the `validate` branch as well since we are pushing and working a lot with it currently

for reference please check [this PR merged into develop](https://github.com/C-Accel-CRIPT/Python-SDK/pull/130)


nh916:
closing this PR because I think it is already within the develop branch

https://github.com/C-Accel-CRIPT/Python-SDK/tree/develop/.github/workflows

PR #132, opened 2023-05-31
Feature: `File.upload()` & `File.download()`

nh916:
# Description

## Changes
* wrote `_file_source_setter` with typehints and docstrings
    * this is not my favorite design because it is essentially a setter for the setter, but since both constructor and setter check node schema and the file node might not be valid yet because everything has not been set yet, it might create an issue
        * this avoids duplicate code for both setter and constructor, but it is not the best
    * Since Python says whatever that can be a function should be a function, I wrote `_file_source_setter`  as a function
         * since it does not refer to self it can easily be made into a function for easy refactoring if needed.
* updated docs for file source in constructor
    > Note: File constructor is longer than expected. Some of it can later be cut down to be more concise with user testing or just review
* quickly updated `API.upload_file()` while I was on it to finish it off
* importing cached API into both upload and download function/method separately to avoid `circular import error`

## Tests
* write test for file source constructor setter and it passed successfully
* wrote test for file source setter and it passed successfully

## Notes
* we could give users the ability to name their file whatever name they want or take it from the file node name, but I wasn't sure how much that was needed so for now I am just taking the name from the file node and we can add that as an upgrade if needed later
* Not sure if the design is good or not, but users can let us know in user testing

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -626,7 +626,6 @@ class API:
         with open(destination_file_path, "wb") as file:
             file.write(file_contents)
 
-    # TODO reset to work with real nodes node_type.node and node_type to be PrimaryNode
     def search(
         self,
         node_type: BaseNode,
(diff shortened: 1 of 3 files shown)


nh916:
This test needs a valid API token to work correctly with the cloud storage provider

InnocentBug (approved):

nh916:
@InnocentBug made a small adjustment to it, I was going to merge it in because I think the change is super small, but I figured it would be best to ask for confirmation just in case before merging it in

InnocentBug (approved):

PR #138, opened 2023-06-02
Create doctest CI

nh916:
# Description
* Create doctest CI check with pytest
* I wanted to use the [python module doctest](https://docs.python.org/3/library/doctest.html#simple-usage-checking-examples-in-docstrings), but it was harder to work with, so I ended up using the [Pytest docstest module](https://docs.pytest.org/en/7.1.x/how-to/doctest.html) instead
    * > We can eventually add both later if we want to, but that might be too many checks
* All code examples, however, would have to change to work with the doctests and for it to be able to find them and work with them. otheriwse it won't be able to find them as they are currently.
    * > Hopefully, there can be someone that can help with documentation for us in the near future
  
```python
# test docstring code that can be found by doctests
hello this is a doctest
>>> x = 3
>>> x
3
```

## Tests
* tested it on [forked repository](https://github.com/nh916/Python-SDK-tests/actions/workflows/pytest_doctest.yaml)

## Note
pytest considers it a test failure if no tests are discovered or executed


nh916:
putting this workflow into the big Pytest Doctest PR: https://github.com/C-Accel-CRIPT/Python-SDK/pull/371

PR #148, opened 2023-06-08
add process back into material

InnocentBug:
# Description
Process is again part of the material node.

## Changes
I think it used to be an array, now it is a single value.
## Tests
Testing this like ususal.
## Known Issues
None
## Checklist

Not sure about documentation, the old one with the process was still in.

### DIFF
--- a/src/cript/nodes/primary_nodes/material.py
+++ b/src/cript/nodes/primary_nodes/material.py
@@ -1,7 +1,8 @@
 from dataclasses import dataclass, field, replace
-from typing import Any, List
+from typing import Any, List, Union
 
 from cript.nodes.primary_nodes.primary_base_node import PrimaryBaseNode
+from cript.nodes.primary_nodes.process import Process
 
 
 class Material(PrimaryBaseNode):
@@ -71,6 +72,7 @@ class Material(PrimaryBaseNode):
         identifiers: List[dict[str, str]] = field(default_factory=dict)
         # TODO add proper typing in future, using Any for now to avoid circular import error
         component: List["Material"] = field(default_factory=list)
+        process: Union[Process, None] = None
         property: List[Any] = field(default_factory=list)
         parent_material: List["Material"] = field(default_factory=list)
         computational_forcefield: List[Any] = field(default_factory=list)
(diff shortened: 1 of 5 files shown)


nh916:
@InnocentBug after we get all of the PR in and everything is working, I think it might end up being a good idea if we add a rule to our GitHub repository to only allow merges if all tests pass, that way we won't have to worry about us merging in a branch that is faulty. what do you think?

InnocentBug:
> @InnocentBug after we get all of the PR in and everything is working, I think it might end up being a good idea if we add a rule to our GitHub repository to only allow merges if all tests pass, that way we won't have to worry about us merging in a branch that is faulty. what do you think?

well, yes and no.

If you set this up in github, trunk merge q won't work.

But if we merge everything with the Q exclusively, this is taken care of automatically.

src/cript/nodes/primary_nodes/material.py
```diff
@@ -71,6 +72,7 @@ class JsonAttributes(PrimaryBaseNode.JsonAttributes):
         identifiers: List[dict[str, str]] = field(default_factory=dict)
         # TODO add proper typing in future, using Any for now to avoid circular import error
         component: List["Material"] = field(default_factory=list)
+        process: Union[Process, None] = None
```
nh916, line 75: ooh for this you want to use [Optional](https://docs.python.org/3/library/typing.html#typing.Optional)

`Optional` is pretty much equal to `Union[something, None]` 
> I have this in a bunch of places that I should remove later because `Union` is supposed to be used only if it is `" you can pass either this or that"` `Union[object1, object2]` instead of `Optional` if something can be there or not

leaned all about typing when I was working with mypy!

nh916 (approved):

src/cript/nodes/primary_nodes/material.py
nh916: if you submit another PR changing this, I can quickly approve it!

InnocentBug: sure, happy to learn here.

This is done.

nh916 (approved):

nh916:
> > @InnocentBug after we get all of the PR in and everything is working, I think it might end up being a good idea if we add a rule to our GitHub repository to only allow merges if all tests pass, that way we won't have to worry about us merging in a branch that is faulty. what do you think?
> 
> well, yes and no.
> 
> If you set this up in github, trunk merge q won't work.
> 
> But if we merge everything with the Q exclusively, this is taken care of automatically.

okay got it!

PR #142, opened 2023-06-07
made progress on fixing mypy static type hinting errors

nh916:
# Description
* fixing any static type hinting errors with mypy
* this could help us catch errors and be an added layer of precaution

## Known Issues
* The code looks a bit uglier and harder to read
* fixed most errors, but there are still a lot of errors that remain

PR #151, opened 2023-06-08
Create `mypy CI` & `Fix mypy typing` errors & add `@beartype`

nh916:
# Description
redoing PR because a lot of merge conflicts from original
closed PR: https://github.com/C-Accel-CRIPT/Python-SDK/pull/142 and using this instead

## Changes
* changed many typings to use `Optional` for all nodes

## Checklist


nh916:
@InnocentBug I noticed some of the mypy errors are when we are returning a copy of a list, but the list might be None and that could throw an error. I told it to ignore it for now, but we may need to address it later

InnocentBug:
> @InnocentBug I noticed some of the mypy errors are when we are returning a copy of a list, but the list might be None and that could throw an error. I told it to ignore it for now, but we may need to address it later

Wait, the list should never be a `None`.
It should be an empty list then.

We only briefly allow `None` for lists in the `__init__`, but that gets converted into an empty list immediately.
The acutall `json_attrs` should never contain a `None`, hence shouldn't be an issue.

nh916:
> > @InnocentBug I noticed some of the mypy errors are when we are returning a copy of a list, but the list might be None and that could throw an error. I told it to ignore it for now, but we may need to address it later
> 
> Wait, the list should never be a `None`. It should be an empty list then.
> 
> We only briefly allow `None` for lists in the `__init__`, but that gets converted into an empty list immediately. The acutall `json_attrs` should never contain a `None`, hence shouldn't be an issue.

I remember in the past when I was working with it I was getting empty list errors with `list.copy`, but I think that was because I didn't have empty list conversions from None in the constructor. 

I think we should be able to safely ignore this typing error for now.

nh916:
@InnocentBug not sure how to fix the `quantity.py` type errors 

we have value and uncertainty as optional, but then later in the code when using getters we are saying that we will always return a number, when it could also be None

```python
        value: Optional[Number] = None
        uncertainty: Optional[Number] = None
```

mypy is right, this logic doesn't make sense. However, at the same time we are requiring a number in constructor and setters, so I think it shouldn't ever end up being `None`

I'm thinking of telling mypy to ignore the issue. What are your thoughts?

nh916:
this is as far as I got 

```bash
src\cript\nodes\cache.py:5: error: Need type annotation for "_node_cache"  [var-annotated]
src\cript\nodes\core.py:212: error: Incompatible default for argument "handled_ids" (default has type "None", argument has type "Set[Any]")  [assignment]
src\cript\nodes\core.py:212: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
src\cript\nodes\core.py:212: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
src\cript\nodes\core.py:257: error: Argument 2 to "CRIPTJsonSerializationError" has incompatible type "JsonAttributes"; expected "str"  [arg-type]
src\cript\nodes\subobjects\quantity.py:142: error: Incompatible return value type (got "Optional[Number]", expected "Union[int, float, str]")  [return-value]
src\cript\nodes\subobjects\quantity.py:183: error: Incompatible return value type (got "Optional[Number]", expected "Number")  [return-value]
src\cript\api\api.py:472: error: Value of type "Response" is not indexable  [index]
src\cript\api\api.py:473: error: Value of type "Response" is not indexable  [index]
src\cript\api\api.py:524: error: Name "api_endpoint" already defined on line 521  [no-redef]
src\cript\api\api.py:527: error: Name "api_endpoint" already defined on line 521  [no-redef]
src\cript\api\api.py:530: error: Name "api_endpoint" already defined on line 521  [no-redef]
src\cript\nodes\util\__init__.py:23: error: Need type annotation for "handled_ids" (hint: "handled_ids: Set[<type>] = ...")  [var-annotated]
src\cript\nodes\util\__init__.py:24: error: Need type annotation for "condense_to_uuid" (hint: "condense_to_uuid: List[<type>] = ...")  [var-annotated]
src\cript\nodes\util\__init__.py:221: error: Need type annotation for "node_dict" (hint: "node_dict: Dict[<type>, <type>] = ...")  [var-annotated]
src\cript\nodes\util\__init__.py:221: error: Argument 1 to "dict" has incompatible type "str"; expected "Iterable[Tuple[<nothing>, <nothing>]]"  [arg-type]
src\cript\nodes\util\__init__.py:221: note: Following member(s) of "str" have conflicts:
src\cript\nodes\util\__init__.py:221: note:     Expected:
src\cript\nodes\util\__init__.py:221: note:         def __iter__(self) -> Iterator[Tuple[<nothing>, <nothing>]]
src\cript\nodes\util\__init__.py:221: note:     Got:
src\cript\nodes\util\__init__.py:221: note:         def __iter__(self) -> Iterator[str]
src\cript\nodes\util\__init__.py:250: error: Argument "object_hook" to "loads" has incompatible type "Callable[[str], Dict[Any, Any]]"; expected "Optional[Callable[[Dict[Any, Any]], Any]]"  [arg-type]
Found 15 errors in 5 files (checked 44 source files)
```

these errors, I am unsure of how to fix currently

nh916:
@InnocentBug I think when I was switching between branches `@beartype` came along for the ride without me noticing and is now in this branch when I was originally going to put it into a separate PR. I think, I'm just going to embrace it and just add beartype here too

nh916:
mypy is so sweet to work with!

nh916:
@InnocentBug can you trunk format this PR, I tried doing it a few times and it kept not working

InnocentBug (dismissed):

InnocentBug (approved):

nh916:
@InnocentBug I noticed you approved this PR, but the tests are failing. Are you thinking that we should input the PR into develop and then fix the tests later?

nh916:
converting to draft until beartype issues are fixed

requirements_dev.txt
```diff
@@ -2,4 +2,6 @@
 black==23.3.0
 mypy==1.3.0
 pytest-cov==4.1.0
-coverage==7.2.3
\ No newline at end of file
+coverage==7.2.7
+types-jsonschema>=4.17
+types-requests>=2.31
```
nh916, line 7: lets make these into exact versions just in case, so we can eliminate one potential place for a bug

src/cript/api/api.py
```diff
@@ -463,16 +470,13 @@ def save(self, project: Project) -> None:
...
-        # TODO work on this later to allow for PATCH as well
-        response = requests.post(url=f"{self._host}/{project.node_type.lower()}", headers=self._http_headers, data=project.json)
+        response: Dict = requests.post(url=f"{self._host}/{project.node_type.lower()}", headers=self._http_headers, data=project.json).json()
 
-        response = response.json()
-
-        # if htt response is not 200 then show the API error to the user
+        # if http response is not 200 then show the API error to the user
         if response["code"] != 200:
             raise CRIPTAPISaveError(api_host_domain=self._host, http_code=response["code"], api_response=response["error"])
```
nh916, line 477: I think the way I originally did this works, but a better/cleaner way for the future might be to use the requests function to do this

```python
    response.raise_for_status()  # Raises an HTTPError if the status code is not 200
````

but not fully sure about it, it needs more research and testing to be sure it works correctly and the error it gives are good

File: src/cript/api/api.py [hunk 3]
```diff
@@ -516,20 +520,22 @@ def search(
...
-            api_endpoint: str = f"{self._host}/search/exact/{node_type}"
+            api_endpoint = f"{self._host}/search/exact/{node_type}"
 
         elif search_mode == SearchModes.UUID:
-            api_endpoint: str = f"{self._host}/{node_type}/{value_to_search}"
+            api_endpoint = f"{self._host}/{node_type}/{value_to_search}"
             # putting the value_to_search in the URL instead of a query
             value_to_search = None
 
+        assert api_endpoint != ""
```
nh916, line 539: I like the way that this is handled, I doubt it'll ever be empty, but we should probably have something for the user in case they bump into an `assertion error` and have no idea where it came from and how to fix it

src/cript/api/paginator.py
```diff
@@ -123,6 +126,7 @@ def current_page_number(self) -> int:
         return self._current_page_number
 
     @current_page_number.setter
+    @beartype
```
nh916, line 129: we have to at some point write a wiki about beartype, why we used it, what it does, why we put it specifically always under the `property` decorator, so that anyone maintaining/helping with it will not get themselves into trouble later and create issues

InnocentBug (approved):

src/cript/nodes/core.py
```diff
@@ -150,6 +150,8 @@ def validate(self, api=None) -> None:
 
     @classmethod
     def _from_json(cls, json_dict: dict):
+        # TODO find a way to handle uuid nodes only
```
nh916, line 153: this is me being picky, I would say if you can make this more descriptive so anyone looking at it knows what you are talking about and can know to work around it or possibly help with it. 

If it would be too long to write out then it is no big deal either, it is just a comment

src/cript/nodes/primary_nodes/computation.py
```diff
@@ -142,6 +145,7 @@ def __init__(
     # ------------------ Properties ------------------
```
nh916, line 145: I wonder if it is a good idea to keep this here, or if it is unneeded and should be removed. I can see both sides to it, and it is no big deal at all either way, but always looking out for improvements

requirements_dev.txt
nh916: updated this!

src/cript/api/api.py
InnocentBug: yeah, something for later to consider.

src/cript/api/paginator.py
InnocentBug: yeah, that makes sense. if we stick with beartype

src/cript/nodes/primary_nodes/computation.py
InnocentBug: I am neutral on this.

src/cript/nodes/core.py
InnocentBug: yeah, maybe. I am not sure if I want it even there.

The new JSON hook class might be better for this.

src/cript/api/api.py
InnocentBug: we might want to write somethingin the docs, that if they ever encounter an assertion error, it is a bug and they should report it.

InnocentBug (approved):

src/cript/api/api.py
nh916: I think putting it within docs can be good, but I bet we can put it as a good error, or just never hit this case. Something we can do for later

src/cript/nodes/core.py
nh916: hmm, that is a good point.

Before upgrading the serializer and reserializer lets be sure to walk through it a bit because if we end up switching to Pydantic it could end up doing a lot of the heavy lifting for us

src/cript/nodes/primary_nodes/computation.py
nh916: yeah same

PR #159, opened 2023-06-09
Update simple-issue.md template

nh916:
# Description
updating the template with things that were wrong. Very simple update

## Changes
* changing `Description` to `h1`
* removing title because it always auto fills and it is annoying to have to remove
    * just replacing it with something empty


InnocentBug (approved):
Trunk doesn’t like that change though

nh916:
@InnocentBug the prettier issue, I understand and can fix easily, however, for the `multiple headings` I think it thinks the yaml at the top `name` is another `h1` heading and it isn't. I think we can add a config file with `.markdownlint.json` and tell it to ignore yaml while liniting md files

```json
{
  "MD025": false
}
```

PR #163, opened 2023-06-13
update synthesis example from old SDK

InnocentBug:
# Description
Adding the synthesis example for the old SDK, now working with the new SDK.

## Changes
properties can have non-number values
## Known Issues
Not integrated into docs.
No links to anything included and not highlighted in Readme.

## Checklist


InnocentBug:
This is just a start and something to iterate upon.

nh916:
I tried a few things on this, I think we might have to put `examples` inside of `docs/` because otherwise it is inaccessible to `mkdocs.yaml`

src/cript/__init__.py
```diff
@@ -1,4 +1,12 @@
 # trunk-ignore-all(ruff/F401)
+# trunk-ignore-all(ruff/E402)
+
+# TODO fix beartype warning for real
+from warnings import filterwarnings
+
+from beartype.roar import BeartypeDecorHintPep585DeprecationWarning
+
+filterwarnings("ignore", category=BeartypeDecorHintPep585DeprecationWarning)
```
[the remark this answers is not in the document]
InnocentBug: yes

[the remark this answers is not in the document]
nh916: well done!

nh916:
```
To connect to CRIPT, you must enter a `host` and an `API Token`. 
For most users, `host` will be `https://criptapp.org`.
Here we assume that your API key is stored in an environment variable called `CRIPT_TOKEN`.
You can export your token like this by executing `export CRIPT_TOKEN="my token text"` 
on the command line for mac or linux, more permanently, 
you can add this line to `~/.bashrc` if you are using `bash` as a shell.
```

I would argue this is too technical and can be made simpler for demonstration for the user and under it we put a warning that says use environment variables. I'm thinking something like this

image

^ the above warning box is a bit too wordy and needs to be trimmed down, but unsure of how to currently

> also for documentation I would suggest always using named arguments to make the code clearer to the user of what it is because this will be their first time seeing it

I'm putting all the changes that I make in a PR to merge into this branch

examples/synthesis.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+These materials are simple, notice how we use the SMILES notation here as an identifier for the material.
+Similarly, we can create more initial materials.
+
+```python
+toluene = cript.Material(name="toluene", identifiers=[{"smiles": "Cc1ccccc1"}])
+styrene = cript.Material(name="styrene", identifiers=[{"smiles": "c1ccccc1C=C"}])
+butanol = cript.Material(name="1-butanol", identifiers=[{"smiles": "OCCCC"}])
+methanol = cript.Material(name="methanol", identifiers=[{"smiles": "CO"}])
```
nh916, line 103: if possible see if you can have a different identifier for each material so we can show off all of our different identifiers

```
bigsmiles,
smiles,
pubchem_cid,
chemical_id,
chem_repeat,
````

I would pick whichever that is most popular, so users know that we are compatible with their favorite way of identifying materials and maybe something that CRIPT specifically does, so they know there are more possibilities and the things they've wanted to do already exists and they can start using CRIPT for it

File: examples/synthesis.md [hunk 3]
```diff
@@ -0,0 +1,228 @@
...
+```python
+process = cript.Process(
+    name="Anionic of Synthesis Poly-Styrene",
+    type="multistep",
+    description="In an argon filled glove box, a round bottom flask was filled with 216 ml of dried toluene. The "
+    "solution of secBuLi (3 ml, 3.9 mmol) was added next, followed by styrene (22.3 g, 176 mmol) to "
+    "initiate the polymerization. The reaction mixture immediately turned orange. After 30 min, "
+    "the reaction was quenched with the addition of 3 ml of methanol. The polymer was isolated by "
+    "precipitation in methanol 3 times and dried under vacuum.",
+)
```
nh916, line 127: a bit too long, but I think it's okay for now

examples/synthesis.md
```diff
@@ -0,0 +1,228 @@
...
+Now we can save the project to CRIPT via the api object.
+
+```python
+project.validate()
+print(project.get_json(indent=2).json)
+# api.save(project)
+```
+
+```python
+# Don't forget to disconnect once everything is done
```
nh916, line 226: I really like this comment for some reason. It just makes it feel so human and nice!

examples/synthesis.md
```diff
@@ -0,0 +1,228 @@
...
+
+```python
+import cript
+
+with cript.API("http://development.api.mycriptapp.org/", None):
+    pass
+```
+
+The `None` arguments signals here, that we read the token from the environment as discussed above.
+You may notice, that we are not executing any code inside the context manager block. If you were to write a python script, compared to a jupyter notebook, you would add all of the following code inside that block.
```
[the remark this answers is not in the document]
InnocentBug: Yeah, maybe. But using Jupyter notebooks here, is easier to integrate into the documentation.
`jupytext` can convert very easily between the formats, so no problem if we change our mind later.

examples/synthesis.md
```diff
@@ -0,0 +1,228 @@
...
+```python
+print(project.json)
+print("\nOr more pretty\n")
+print(project.get_json(indent=2).json)
+```
+
+# Create an Experiment node
+
+The collection holds a series of `Experiment` nodes.
+Every experiment holds processes and data related this particular experiment.
```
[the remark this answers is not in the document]
InnocentBug: I just didn't know the links, when I wrote it.
You do and I see, that you updated it in your PR.

examples/synthesis.md
InnocentBug: Yes, agreed.
However, the identifiers don't quite make sense for these materials.
All of them should have `smiles`, we can add more to show them off.

Except for bigSMILES, since these are not polymers, bigSMILES and smiles are identical.

examples/synthesis.md
InnocentBug: awesome :)

InnocentBug:
Thanks for helping out and optimizing the documentation here. very much appreciated. and you know your documentation system better then I do.

Also, I already setup a test that runs the notebook to make sure it doesn't fail.

examples/synthesis.md
InnocentBug: I added those.
And actually that was very good, because I used the wrong identifier for the solution.
BTW, if you ever have to work with this: https://cactus.nci.nih.gov/chemical/structure
super helpful to convert between different identifiers.

(Maybe we should even include something, that suggests identifiers, automatically from this service.)

nh916 (dismissed):

InnocentBug:
@nh916  can you double check, I resolved the merge conflict in `mkdocs.yml` correctly?

nh916:
> @nh916 can you double check, I resolved the merge conflict in `mkdocs.yml` correctly?

sure can!

nh916:
seems like everything is working well! If we miss anything for documentation, I am less worried because we can easily fix it and there usually isn't any huge consequences that could happen.

The only thing we have to be careful with is the formatting of the [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) because those are very sensitive and can be broken very easily and quickly. 

I like them and I think that they add some nice color and contrast to the documentation making it pleasing and easier to read but they are a pain because by changing the format slightly breaks them. 

I think if we just use markdown instead our documentation would probably be a bit easier to work with and more portable that we can easily switch between different platforms, but I think that is something to consider for later and it is good as is currently.

.github/workflows/test_examples.yml
```diff
@@ -0,0 +1,35 @@
+name: Test Jupyter Notebook
```
nh916, line 1: I would add some comments to the top of the workflow because I've been noticed that since our CI is growing I am losing track of what does what and where it is applicable. I think it can also be helpful for future developers as well so they can easily understand how to work with it

InnocentBug:
> @InnocentBug seems like everything is working well! If we miss anything for documentation, I am less worried because we can easily fix it and there usually isn't any huge consequences that could happen.
> 
> The only thing we have to be careful with is the formatting of the [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) because those are very sensitive and can be broken very easily and quickly.
> 
> I like them and I think that they add some nice color and contrast to the documentation making it pleasing and easier to read but they are a pain because by changing the format slightly breaks them.
> 
> I think if we just use markdown instead our documentation would probably be a bit easier to work with and more portable that we can easily switch between different platforms, but I think that is something to consider for later and it is good as is currently.

oh, I see. Yeah that is annoying.
We use markdownlint with trunk here and that assumes standard markdown.

I can see how that easily breaks.

.github/workflows/test_examples.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: install test dependency
+        run: python3 -m pip install -r requirements_dev.txt
+      - name: install module
+        run: python3 -m pip install .
+      - name: prepare notebook
+        run: jupytext --to py docs/examples/synthesis.md
+      - name: Run script
+        run: python3 docs/examples/synthesis.py
```
nh916, line 35: this is me being extremely picky, but I like formatting it so there is one line of space between each step, I think that just reads easier, but that is complete personal preference

```yaml
steps:
      - name: Checkout
        uses: actions/checkout@v3

      - name: install test dependency
        run: python3 -m pip install -r requirements_dev.txt

      - name: install module
        run: python3 -m pip install .

      - name: prepare notebook
        run: jupytext --to py docs/examples/synthesis.md

      - name: Run script
        run: python3 docs/examples/synthesis.py
```

.trunk/configs/.markdownlint.yaml
```diff
@@ -9,3 +9,4 @@ spaces: false
 url: false
 whitespace: false
 MD041: false
+MD046: false
```
nh916, line 12: This is really nice that you figure out all of this! I was struggling trying to figure it out

nh916 (approved):

examples/synthesis.md
nh916: That website is super helpful! I will save it somewhere. I wonder if it would be a good idea to put it within our `#cript-dev` channel and pin it so everyone has access to it for future development, or just put it within our wiki to use for later

nh916:
> > @InnocentBug seems like everything is working well! If we miss anything for documentation, I am less worried because we can easily fix it and there usually isn't any huge consequences that could happen.
> > The only thing we have to be careful with is the formatting of the [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) because those are very sensitive and can be broken very easily and quickly.
> > I like them and I think that they add some nice color and contrast to the documentation making it pleasing and easier to read but they are a pain because by changing the format slightly breaks them.
> > I think if we just use markdown instead our documentation would probably be a bit easier to work with and more portable that we can easily switch between different platforms, but I think that is something to consider for later and it is good as is currently.
> 
> oh, I see. Yeah that is annoying. We use markdownlint with trunk here and that assumes standard markdown.
> 
> I can see how that easily breaks. Is there a way, to have the doc tests breaking when that happens? That would be handy to check then.

---

I am not totally sure because if we don't format it correctly the program can just think that it is valid markdown and it would work, but maybe the documentation test build CI would complain about it because something else could also break indirectly?

Not sure about this currently, I'm thinking of coming back to it later and writing an issue about it for now maybe

nh916 (approved):

nh916:
@InnocentBug if you want a sanity check and want me to review the docs feel free to let me know at any time

nh916:
I also wrote a [wiki documentation for our documentation](https://github.com/C-Accel-CRIPT/Python-SDK/wiki/Documentation) that if you are having a hard time understanding the documentation structure or anything else, that could come in handy. Feel free to edit it if I missed anything or let me know to add more/improve it if needed

docs/examples/synthesis.md
```diff
@@ -0,0 +1,291 @@
...
+You may notice, that we are not executing any code inside the context manager block.
+If you were to write a python script, compared to a jupyter notebook, you would add all the following code inside that block.
+Here in a jupyter notebook, we need to connect manually. We just have to remember to disconnect at the end.
+
+```python
+api = cript.API("http://development.api.mycriptapp.org/", None)
+api = api.connect()
+```
+
+# Create a Project
```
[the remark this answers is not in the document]
nh916: fixed it!

nh916 (approved):

PR #164, opened 2023-06-13
optimize GitHub CI

nh916:
# Description
optimized GitHub workflow CI a bit. Removed some tests that we had that were running on all OS and on different versions of Python. Many of them will give the same answer regardless of Python version and OS, so they have been changed to only Ubuntu and Python 3.11 for speed.

I think this should be relatively safe, and if anything is wrong we can change it pretty easily

## Checklist


InnocentBug (approved):

InnocentBug:
we could consider remove macos as well.

nh916:
test coverage still taking long

nh916:
for `macos` jobs it seems like it is taking longest to just get the OS

PR #168, opened 2023-06-14
WIP: Writing Integration Tests

nh916:
# Description
Write integration tests and write all necessary code to work with or around the API for the SDK to integrate with it.

### Writing Integration Test
* wrote an` Create` and `Read` integration test function for all nodes, sub-objects, and supporting nodes
    * > skipped `User` node for now because not sure if we'll need have POST and GET integrations for it, and the other nodes are more crucial right now
    * > skipping `Update` and `Delete` parts of the `CRUD `operation test until later when those are available
    * Every integration test is essentially the shortest distance (or what makes the most sense) from the node that needs integration test to the `Project` node
        * Saving the project to the API
            * every integration test has a unique name with a UUID in the name to avoid any duplication
         * then getting the node by name with `cript.API.search` and `SearchModes.EXACT_SEARCH`
         * comparing the JSON from SDK and API to be sure that they are equal
             * > comparing the JSON instead of the Python node objects because it is easier to compare nested things
         * then we can be sure that the Project was able to be successfully saved and gotten from the API
* Comparing the JSON that was originally created and the one that was deserialized was challenging and here is what I did
    * When I send JSON to the API and when the API sends back JSON, they are different JSONs that send back more information. 
    * When deserializing it I cannot just compare them one by one because they are in different order, and some have full nodes and others have just UUID references. 
    * I think the code that I have basically compares the fields that they have in common and just ignores the UID field because that will always be different with API and SDK.
    * This is not the best strategy because if there is a discrepancy in keys they don't have in common it might create issues that we may not catch
       * this can be upgraded in the future, but for now I think it is okay
### Bad UUID API Error
One of the errors that we encountered was a “Bad UUID Error” from the API. This error essentially meant that we have to figure out which nodes already exist in the database or not. Some places in the JSON the API requires us to not send the full node, but to only send an ID to refer to the node, however, it requires us to send a UUID if the node already exists in the database or UID if it needs to be created, and we do not have a way of knowing if the node already exists in the database or not without checking every single node with the API.

The way we fixed this was essentially that we send our JSON as usual, then we check the response, and if it has `Bad UUID Error` then we find the UUID that was the issue, send that to the API, then when it is created, send the JSON again. Repeat this process until the data has been uploaded to API.

This was the most efficient and fastest way that we could figure out.

### Deserialization Errors and Bad API Output
There were some places that the API expected an input of `int` for a value, but then gave an output of string for that same value. We fixed this within our nodes by allowing for either a `string` or `integer`

This is inconsistent with the data model, but we'll have to accrue tech debt for now, and we can come back and fix it maybe in the future.

### Fixture Upgrades
* updated `complex_property_node`
     * changed instantiation of the node to named arguments
     * changing notes from `"notes"` to `"my complex_property_node notes"` to easily know that the notes are coming through correctly
          * > notes is too generic and if we have "notes" for any other node we might not be able to catch if the notes are working correctly through our code or if they are being swapped and changed without us noticing
     * renamed test variable to make reading it easier
     * added small docstring to the function fixture
* updated `complex_equipment_node`
    * added docstrings
    * made it into named arguements during instantiation
    * renamed the variable from `e` to `my_complex_equipment`
    * changed description a bit to help identify it if needed in tests
* updated `complex_computational_forcefield_node` fixture
    * changed it to named arguments
   * changed the variable name
   * added to description to easily identify it if needed
   * added minimal docstrings to it
* updated `complex_software_configuration_node` fixture
    * changed it to named arguments
    * changed the variable name
    * added to notes to easily identify it if needed
    * added minimal docstrings to it
 * updated `complex_parameter_node` fixture
    * changed it to named arguments
    * added minimal docstrings to it
* updated `complex_algorithm_node` fixture
    * changed it to named arguments
    * added minimal docstrings to it
* upgraded `complex_citation_node` fixture
    * made it into named arguments during instantiation
    * added minimal docstrings to it
* wrote `simple_condition_node` fixture for easier tests, so that way I can isolate on the thing I am testing without any noise from any other node

## Known Issues
* The code is becoming less clean and harder to read. We must address this soon after the ACS convention.
* Refactoring must happen very soon
* When saving a collection with citation with reference, the API says it has been saved fine, but when getting the collection, it shows that it does not have a Citation sub-object on the Collection
  * We wrote trello tickets for them here and are waiting for the API to fix that issue:
      * https://trello.com/c/QkUTlD5j
      * https://trello.com/c/QkUTlD5j

## Attributions
Big thanks to @InnocentBug for helping with deserialization errors, bad uuid errors, and brainstorm problems with me,

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,7 +33,8 @@ jobs:
           export CRIPT_TOKEN="125433546"
           export CRIPT_HOST="http://development.api.mycriptapp.org/"
           python3 -m pip install pytest
-          python3 -m pip install -r requirements.txt
+          python3 -m pip install -r requirements.txt 
+          python3 -m pip install -r requirements_dev.txt 
           python3 -c "import cript"
           export CRIPT_TOKEN="125433546"
           export CRIPT_HOST="http://development.api.mycriptapp.org/"
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -7,3 +7,4 @@ types-jsonschema==4.17.0.8
 types-requests==2.31.0.1
 types-boto3==1.0.2
 deepdiff==6.3.0
+jupytext==1.13.6
--- a/src/cript/nodes/subobjects/condition.py
+++ b/src/cript/nodes/subobjects/condition.py
@@ -80,12 +80,12 @@ class Condition(UUIDBaseNode):
         key: str = ""
         type: str = ""
         descriptor: str = ""
-        value: Union[Number, None] = None
+        value: Optional[Union[Number, str]] = None
         unit: str = ""
-        uncertainty: Union[Number, None] = None
+        uncertainty: Optional[Union[Number, str]] = None
         uncertainty_type: str = ""
-        set_id: Union[int, None] = None
-        measurement_id: Union[int, None] = None
+        set_id: Optional[int] = None
+        measurement_id: Optional[int] = None
         data: List[Data] = field(default_factory=list)
 
     _json_attrs: JsonAttributes = JsonAttributes()
@@ -95,13 +95,13 @@ class Condition(UUIDBaseNode):
         self,
         key: str,
         type: str,
-        value: Number,
+        value: Union[Number, str],
         unit: str = "",
         descriptor: str = "",
-        uncertainty: Union[Number, None] = None,
+        uncertainty: Optional[Union[Number, str]] = None,
         uncertainty_type: str = "",
-        set_id: Union[int, None] = None,
-        measurement_id: Union[int, None] = None,
+        set_id: Optional[int] = None,
+        measurement_id: Optional[int] = None,
         data: Optional[List[Data]] = None,
         **kwargs
     ):
@@ -285,7 +285,7 @@ class Condition(UUIDBaseNode):
 
     @property
     @beartype
-    def value(self) -> Union[Number, None]:
+    def value(self) -> Optional[Union[Number, str]]:
         """
         value or quantity
 
@@ -302,7 +302,7 @@ class Condition(UUIDBaseNode):
         """
         return self._json_attrs.value
 
-    def set_value(self, new_value: Number, new_unit: str) -> None:
+    def set_value(self, new_value: Union[Number, str], new_unit: str) -> None:
         """
         set the value for this Condition subobject
 
@@ -347,7 +347,7 @@ class Condition(UUIDBaseNode):
 
     @property
     @beartype
-    def uncertainty(self) -> Union[Number, None]:
+    def uncertainty(self) -> Optional[Union[Number, str]]:
         """
         set uncertainty value for this Condition subobject
 
@@ -365,7 +365,7 @@ class Condition(UUIDBaseNode):
         return self._json_attrs.uncertainty
 
     @beartype
-    def set_uncertainty(self, new_uncertainty: Number, new_uncertainty_type: str) -> None:
+    def set_uncertainty(self, new_uncertainty: Union[Number, str], new_uncertainty_type: str) -> None:
         """
         set uncertainty and uncertainty type
 
--- a/.github/ISSUE_TEMPLATE/simple-issue.md
+++ b/.github/ISSUE_TEMPLATE/simple-issue.md
@@ -1,7 +1,7 @@
 ---
 name: Simple Issue
 about: Describe the issue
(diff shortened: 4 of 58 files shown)


src/cript/nodes/subobjects/condition.py
```diff
@@ -285,7 +285,7 @@ def descriptor(self, new_descriptor: str) -> None:
 
     @property
     @beartype
-    def value(self) -> Union[Number, None]:
+    def value(self) -> Optional[Union[Number, str]]:
```
nh916, line 288: I would use `Union[Number, str, None` here just because it would easier to read in my opinion. When things start to get nested, then I start to have a hard time following it

nh916:
Big thanks to @InnocentBug for helping get all the tests to pass for integration!

nh916:
all integration tests work locally except for `collection.citation.reference` which is an API error documented in this trello ticket: https://trello.com/c/kct2H9EV

the rest we got working

nh916:
waiting for this [PR](https://github.com/C-Accel-CRIPT/Python-SDK/pull/190) to be merged and then will request a review

nh916:
the way we are saving to the API for sure needs a wiki page explaining how we did it

nh916:
I think the reason that test coverage is not 90%+ is because there is a lot of failed tests due to integration tests failing because they don't have a token

requirements_dev.txt
```diff
@@ -6,4 +6,5 @@ coverage==7.2.7
 types-jsonschema==4.17.0.8
 types-requests==2.31.0.1
 types-boto3==1.0.2
+deepdiff==6.3.0
 jupytext==1.13.6
```
nh916, line 10: @InnocentBug is `jupytext` for our documentation?

InnocentBug: yes, and no.
We use jupytext to convert the example `md` documents into python documents and run them.

To ensure, that the examples always work.

nh916:
@InnocentBug 

> yes, and no.
We use jupytext to convert the example md documents into python documents and run them.

> To ensure, that the examples always work.

in that case I'd recommend putting it in `requirementst_docs.txt` because I've put anything that is for the documentation in that file

nh916:
trunk is finding a lot of errors with words that we've already told it to ignore

InnocentBug:
> > 
> 
> in that case I'd recommend putting it in `requirementst_docs.txt` because I've put anything that is for the documentation in that file

I don't think that makes sense.
We load `requirement_devs.txt` to run tests.

InnocentBug:
> trunk is finding a lot of errors with words that we've already told it to ignore

yeah, because the `cspell` config file was broken.

nh916:
> > trunk is finding a lot of errors with words that we've already told it to ignore
> 
> yeah, because the `cspell` config file was broken.

dang that was a great catch! I should've seen the missing comma, can't believe I missed it

nh916:
@InnocentBug big thanks for fixing the import, I just saw that you did that!

InnocentBug (approved):
This looks good to me.

It's a bummer we can automatically test the integration yet.

We may want to wait for #193 to also support `patch`.
But I am OK to wait for this as well.
The patch also doesn't have proper testing yet.

nh916:
> This looks good to me.
> 
> It's a bummer we can automatically test the integration yet.
> 
> We may want to wait for #193 to also support `patch`. But I am OK to wait for this as well. The patch also doesn't have proper testing yet.

I'm thinking of merging this in so we have `create` for now and it is in our develop branch
then branch off of your update branch and start working on the tests while you're working on the code, so we knock it out in parallel

yeah I do agree that it is a bummer that we cannot automatically check integration tests, but it is out of my hands and nothing I can do about it

Let me know if you want me to wait and not merge this in yet

.github/workflows/tests.yml
```diff
@@ -33,7 +33,8 @@ jobs:
           export CRIPT_TOKEN="125433546"
           export CRIPT_HOST="http://development.api.mycriptapp.org/"
           python3 -m pip install pytest
-          python3 -m pip install -r requirements.txt
+          python3 -m pip install -r requirements.txt 
+          python3 -m pip install -r requirements_dev.txt
```
nh916, line 37: we could just do `requirements_dev.txt` because that by default installs `requirements.txt`, but I think it's fine either way

requirements_dev.txt
nh916: so yeah that's for documentation

PR #175, opened 2023-06-22
postpone file update until project save

InnocentBug:
# Description
This postpones the file upload until the save of the project.
I think that is a little more in sync with the user expectation.

## Changes

Also post pone s3 client init and pretreat token.
This helps with the test.
(Before there were situations, where you commented the init of the s3 client out to get the tests passing.)
Now tests should pass, but the client still gets initiated if needed.

## Known Issues

None, yet.

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -63,7 +63,7 @@ class API:
     _COGNITO_LOGIN_PROVIDER: str = "cognito-idp.us-east-1.amazonaws.com/us-east-1_VinmyZ0zW"
     _BUCKET_NAME: str = "cript-development-user-data"
     _BUCKET_DIRECTORY_NAME: str = "python_sdk_files"
-    __s3_client: Any = None  # type: ignore
+    _internal_s3_client: Any = None  # type: ignore
     # trunk-ignore-end(cspell)
 
     @beartype
@@ -208,11 +208,12 @@ class API:
 
         return s3_client
 
+    # Use a property to ensure delayed init of s3_client
     @property
     def _s3_client(self):
-        if self.__s3_client is None:
-            self.__s3_client = self._get_s3_client()
-        return self.__s3_client
+        if self._internal_s3_client is None:
+            self._internal_s3_client = self._get_s3_client()
+        return self._internal_s3_client
 
     def __enter__(self):
         self.connect()
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -63,7 +63,7 @@ class API:
     _COGNITO_LOGIN_PROVIDER: str = "cognito-idp.us-east-1.amazonaws.com/us-east-1_VinmyZ0zW"
     _BUCKET_NAME: str = "cript-development-user-data"
     _BUCKET_DIRECTORY_NAME: str = "python_sdk_files"
-    _s3_client: Any = None  # type: ignore
+    __s3_client: Any = None  # type: ignore
     # trunk-ignore-end(cspell)
 
     @beartype
@@ -158,9 +158,6 @@ class API:
         # TODO might need to add Bearer to it or check for it
         self._http_headers = {"Authorization": f"{self._token}", "Content-Type": "application/json"}
 
-        # TODO commenting this out for now because there is no GitHub API container, and all tests will fail
-        # self._s3_client = self._get_s3_client()
-
         # check that api can connect to CRIPT with host and token
         self._check_initial_host_connection()
 
@@ -194,7 +191,11 @@ class API:
 
         identity_id = auth.get_id(IdentityPoolId=self._IDENTITY_POOL_ID, Logins={self._COGNITO_LOGIN_PROVIDER: self._token})
 
-        aws_credentials = auth.get_credentials_for_identity(IdentityId=identity_id["IdentityId"], Logins={self._COGNITO_LOGIN_PROVIDER: self._token})
+        aws_token = self._token
+        if len(aws_token.split()) > 1:
+            # Remove ID prefix like 'Bearer'
+            aws_token = aws_token.split()[-1]
+        aws_credentials = auth.get_credentials_for_identity(IdentityId=identity_id["IdentityId"], Logins={self._COGNITO_LOGIN_PROVIDER: aws_token})
 
         aws_credentials = aws_credentials["Credentials"]
 
@@ -206,6 +207,12 @@ class API:
         )
 
         return s3_client
+
+    @property
+    def _s3_client(self):
+        if self.__s3_client is None:
+            self.__s3_client = self._get_s3_client()
+        return self.__s3_client
 
     def __enter__(self):
         self.connect()
@@ -513,6 +520,11 @@ class API:
         None
             Just sends a `POST` or `Patch` request to the API
         """
+
+        # Ensure that all file nodes have uploaded there payload before actual save.
+        for file_node in project.find_children({"node": ["File"]}):
+            file_node.ensure_uploaded()
+
         response: Dict = requests.post(url=f"{self._host}/{project.node_type.lower()}", headers=self._http_headers, data=project.json).json()
 
         # if http response is not 200 then show the API error to the user
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -178,41 +178,33 @@ class API:
 
         return host
 
-    def _get_s3_client(self) -> boto3.client:  # type: ignore
+    # Use a property to ensure delayed init of s3_client
+    @property
+    def _s3_client(self) -> boto3.client:  # type: ignore:
         """
-        create a fully authenticated and ready s3 client
+        creates or returns a fully authenticated and ready s3 client
 
         Returns
         -------
         s3_client: boto3.client
             fully prepared and authenticated s3 client ready to be used throughout the script
         """
-        auth = boto3.client("cognito-identity", region_name=self._REGION_NAME)
 
-        identity_id = auth.get_id(IdentityPoolId=self._IDENTITY_POOL_ID, Logins={self._COGNITO_LOGIN_PROVIDER: self._token})
-
-        aws_token = self._token
-        if len(aws_token.split()) > 1:
-            # Remove ID prefix like 'Bearer'
-            aws_token = aws_token.split()[-1]
-        aws_credentials = auth.get_credentials_for_identity(IdentityId=identity_id["IdentityId"], Logins={self._COGNITO_LOGIN_PROVIDER: aws_token})
-
-        aws_credentials = aws_credentials["Credentials"]
-
-        s3_client = boto3.client(
-            "s3",
-            aws_access_key_id=aws_credentials["AccessKeyId"],
-            aws_secret_access_key=aws_credentials["SecretKey"],
-            aws_session_token=aws_credentials["SessionToken"],
-        )
-
-        return s3_client
-
-    # Use a property to ensure delayed init of s3_client
-    @property
-    def _s3_client(self):
         if self._internal_s3_client is None:
-            self._internal_s3_client = self._get_s3_client()
+            auth = boto3.client("cognito-identity", region_name=self._REGION_NAME)
+            identity_id = auth.get_id(IdentityPoolId=self._IDENTITY_POOL_ID, Logins={self._COGNITO_LOGIN_PROVIDER: self._token})
+            # TODO remove this temporary fix to the token, by getting is from back end.
+            aws_token = self._token.lstrip("Bearer ")
+
+            aws_credentials = auth.get_credentials_for_identity(IdentityId=identity_id["IdentityId"], Logins={self._COGNITO_LOGIN_PROVIDER: aws_token})
+            aws_credentials = aws_credentials["Credentials"]
+            s3_client = boto3.client(
+                "s3",
+                aws_access_key_id=aws_credentials["AccessKeyId"],
+                aws_secret_access_key=aws_credentials["SecretKey"],
+                aws_session_token=aws_credentials["SessionToken"],
+            )
+            self._internal_s3_client = s3_client
         return self._internal_s3_client
 
     def __enter__(self):
@@ -521,6 +513,8 @@ class API:
         None
             Just sends a `POST` or `Patch` request to the API
         """
+
+        project.validate()
 
         # Ensure that all file nodes have uploaded there payload before actual save.
         for file_node in project.find_children({"node": ["File"]}):
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -524,7 +524,7 @@ class API:
 
         # Ensure that all file nodes have uploaded there payload before actual save.
         for file_node in project.find_children({"node": ["File"]}):
-            file_node.ensure_uploaded()
+            file_node.ensure_uploaded(api=self)
 
         response: Dict = requests.post(url=f"{self._host}/{project.node_type.lower()}", headers=self._http_headers, data=project.json).json()
 
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -180,7 +180,7 @@ class API:
 
     # Use a property to ensure delayed init of s3_client
     @property
-    def _s3_client(self) -> boto3.client:  # type: ignore:
+    def _s3_client(self) -> boto3.client:  # type: ignore
         """
         creates or returns a fully authenticated and ready s3 client
 
--- a/src/cript/nodes/supporting_nodes/file.py
+++ b/src/cript/nodes/supporting_nodes/file.py
@@ -179,7 +179,7 @@ class File(PrimaryBaseNode):
         -----------
 
         api: cript.API, optional
-           API obejct that performs the upload.
+           API object that performs the upload.
            If None, the globally cached object is being used.
 
         Examples
--- a/src/cript/nodes/supporting_nodes/file.py
+++ b/src/cript/nodes/supporting_nodes/file.py
@@ -25,7 +25,7 @@ def _is_local_file(file_source: str) -> bool:
         return True
 
 
-def _upload_file_and_get_object_name(source: Union[str, Path]) -> str:
+def _upload_file_and_get_object_name(source: Union[str, Path], api=None) -> str:
     """
     uploads file to cloud storage and returns the file link
 
@@ -50,7 +50,8 @@ def _upload_file_and_get_object_name(source: Union[str, Path]) -> str:
     source = str(source)
 
     if _is_local_file(file_source=source):
-        api = _get_global_cached_api()
+        if api is None:
+            api = _get_global_cached_api()
         object_name = api.upload_file(file_path=source)
         # always getting a string for object_name
         source = str(object_name)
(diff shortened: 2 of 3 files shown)


src/cript/api/api.py
```diff
@@ -513,6 +520,11 @@ def save(self, project: Project) -> None:
         None
             Just sends a `POST` or `Patch` request to the API
         """
+
+        # Ensure that all file nodes have uploaded there payload before actual save.
+        for file_node in project.find_children({"node": ["File"]}):
```
nh916, line 525: this is an O(n) operation and `find_children` is an o(n^2) operation, this is total of an o(n^3) operation. This gets me worried that it could be slow very quickly if someone is using our script with something bigger.

How essential would you say it is for upload to happen on save instead of when the file node is created?

src/cript/nodes/supporting_nodes/file.py
```diff
@@ -152,25 +152,26 @@ def __init__(self, name: str, source: str, type: str, extension: str = "", data_
...
         # TODO check if vocabulary is valid or not
         # is_vocab_valid("file type", type)
 
         # setting every attribute except for source, which will be handled via setter
         self._json_attrs = replace(
             self._json_attrs,
             type=type,
-            source=source,
+            # always giving the function the required str regardless if the input `Path` or `str`
+            source=str(source),
```
nh916, line 163: good idea! this is a good catch in case the user gives a `Path` object

src/cript/nodes/supporting_nodes/file.py
```diff
@@ -152,25 +152,26 @@ def __init__(self, name: str, source: str, type: str, extension: str = "", data_
...
             extension=extension,
             data_dictionary=data_dictionary,
         )
 
         self.validate()
 
+    def ensure_uploaded(self):
+        if _is_local_file(file_source=self.source):
+            # upload file source if local file
+            self.source = _upload_file_and_get_object_name(source=self.source)
```
nh916, line 173: This was not obvious to me what it was doing and I had to do some digging around to figure it out. I strongly recommend docstrings.

nh916:
I am not a fan of this design because it essentially puts all the burden of the file upload on the API to loop through the entire tree, find every file node, and upload each file during save. However, then the part that does the actual uploading is in the file node accessing the API cache. This seems a bit backwards to me.

Either put everything for file upload within the API and don't allow anything else to access the `S3 client` because they won't need it or put the upload part within the API and let each file upload their file using the upload method from the API.

With the previous design if the user gets an error they the line number would be very obvious as to what file failed to upload or could not be found in case of wrong path, however, with this design I think the user would not know until later and then it wouldn't even point to the correct line number, and would make debugging harder.

I think even if we want to look into something like this I would put this within the `upgrades` bucket for future where we have more time to test it and mess with it because I want to say that we are getting close to a first working version where we can make a release and then start working on upgrades after that.

src/cript/api/api.py
```diff
@@ -207,6 +208,12 @@ def _get_s3_client(self) -> boto3.client:  # type: ignore
 
         return s3_client
 
+    @property
+    def _s3_client(self):
+        if self.__s3_client is None:
+            self.__s3_client = self._get_s3_client()
+        return self.__s3_client
```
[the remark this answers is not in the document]
nh916: actually the more I think about it, it's not bad because it makes the constructor lighter and this is only initialized when needed. I would like this fully encapsulated into the API, and you have already done that with the underscore prefix. 

I wonder if we need to give good errors for this or not because if this fails out of nowhere the user might be sitting there trying to track down what happened, and AWS boto errors are generally bad, but I think it is mostly fine.

src/cript/api/api.py
```diff
@@ -63,7 +63,7 @@ class API:
     _COGNITO_LOGIN_PROVIDER: str = "cognito-idp.us-east-1.amazonaws.com/us-east-1_VinmyZ0zW"
     _BUCKET_NAME: str = "cript-development-user-data"
     _BUCKET_DIRECTORY_NAME: str = "python_sdk_files"
-    _s3_client: Any = None  # type: ignore
+    __s3_client: Any = None  # type: ignore
```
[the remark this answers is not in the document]
InnocentBug: Not a typo, this is the internal cache with two underscores marking it as even more private as just one.
We can use a different name. Two underscores is usually reserved for builtin names, so thinking about renaming it.

src/cript/api/api.py
```diff
@@ -194,7 +191,11 @@ def _get_s3_client(self) -> boto3.client:  # type: ignore
 
         identity_id = auth.get_id(IdentityPoolId=self._IDENTITY_POOL_ID, Logins={self._COGNITO_LOGIN_PROVIDER: self._token})
 
-        aws_credentials = auth.get_credentials_for_identity(IdentityId=identity_id["IdentityId"], Logins={self._COGNITO_LOGIN_PROVIDER: self._token})
+        aws_token = self._token
+        if len(aws_token.split()) > 1:
+            # Remove ID prefix like 'Bearer'
+            aws_token = aws_token.split()[-1]
```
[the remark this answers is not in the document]
InnocentBug: I am ok, with this being temporary.

It just wasn't working at the moment, so it needed a fix.

I disagree, that `strip` would be better. I thought about it.

Strip only works, if it actually starts with "Bearer", but with how unstable the backend is (especially tokens), I don't want to rely on this. (It used to be a "Token" prefix, and the technology might change again.
(And good luck debugging this from users, since users won't share there token with you.)

The only assumption the split version makes, is that there are no spaces in the token.
And I think that is a very minimal assumption, we have to make.

It also works if there is no prefix, or anything.

You just need no spaces in the token, and the token is last (if there are more sections).

That is why I chose this method over `strip`.

src/cript/api/api.py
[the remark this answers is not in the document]
InnocentBug: Yes, it was broken.

You commented it out entirely, which resulted in `File` nodes being broken.
(It tried to auto upload, but `_s3_client` was `None`.)

So, I needed to find a fix.
And this delayed init, allows us to not break if the `s3` is not used (for example during tests).

But it will automatically work when needed. (Integration tests.)

src/cript/api/api.py
InnocentBug: No, it is not.
First, what do you mean by `n`?

I made a detailed runtime analysis for `find_children` and the runtime is `O(n*(k+m*h)`, with `n` being the number of nodes, `k` being the number of children of the node (assume max. if worst case), `m` being the number of nodes in the searched for attribute, and `h` being the depth of the search dictionary.

I added comments into the code, show this.

Back to this usage: The depth of the search dictionary is 1, we dont' know number of children and length of the attribute lists, but they aren't long. Guaranteed n > m, n > k.

So, the actual `find_children` call is closer to being a linear operation in the number of nodes then to be O(n**2).

And the second loop, is linear in the number of actual file nodes.
(Which you have to be linear in, since we have to upload them all).

So, this is not an issue.

InnocentBug: We could consider rewriting `find_children` as a generator.
In that case, we can eliminate the second loops cost at all.

InnocentBug:
> save. However, then the part that does the actual uploading is in the file node accessing the API cache. This seems a bit backwards to me.

Yes, there was kind of a flaw in my old design.
I improved it. Now a specified API is used to perform the upload of files.
If none is specified the global one is used.

More control.

The logic is this: File objects are in charge of uploading.
They use the API as a tool to do so.
And and during `save` the process ensures that all files are uploaded.

That sounds right to me.

InnocentBug:
> With the previous design if the user gets an error they the line number would be very obvious as to what file failed to upload or could not be found in case of wrong path, however, with this design I think the user would not know until later and then it wouldn't even point to the correct line number, and would make debugging harder.

I can see your logic here, but I still like the new version better.

Suppose you have a user debugging there script.
At the moment, the files are uploaded the second a File node is created.
So if there is a mistake to be debugged after a bunch of File nodes are created, a lot of files are uploaded, but never actually used.
We have no mechanism to prevent uploading a file multiple times in this case.
So we end up with a lot of orphaned uploaded nodes.
This is fine for the users, but really bad for our infrastructure.

The current version, doesn't fully resolve this, but it only occurs if there is an error within `save`.
That should be much better.

And yes, if there is an issue with file uploading, we can improve on the error message, and we could tell the user to manually use `ensure_uploaded` to debug the uploading process for file nodes.

InnocentBug:
> I think even if we want to look into something like this I would put this within the `upgrades` bucket for future where we have more time to test it and mess with it because I want to say that we are getting close to a first working version where we can make a release and then start working on upgrades after that.

No, I need this now.
I am in the situation that I am working and debugging CRIPT scripts.
So the aforementioned situation.
I don't wait for uploading files all the time, plus I don't want to fill the S3 bucket with so many trash files.

nh916:
> > With the previous design if the user gets an error they the line number would be very obvious as to what file failed to upload or could not be found in case of wrong path, however, with this design I think the user would not know until later and then it wouldn't even point to the correct line number, and would make debugging harder.
> 
> I can see your logic here, but I still like the new version better.
> 
> Suppose you have a user debugging there script. At the moment, the files are uploaded the second a File node is created. So if there is a mistake to be debugged after a bunch of File nodes are created, a lot of files are uploaded, but never actually used. We have no mechanism to prevent uploading a file multiple times in this case. So we end up with a lot of orphaned uploaded nodes. This is fine for the users, but really bad for our infrastructure.
> 
> The current version, doesn't fully resolve this, but it only occurs if there is an error within `save`. That should be much better.
> 
> And yes, if there is an issue with file uploading, we can improve on the error message, and we could tell the user to manually use `ensure_uploaded` to debug the uploading process for file nodes.

I do agree on some and disagree on some. 

I completely agree with what your saying about the file uploads. That is very clever and something that I did not see cp,omg. Great Catch! 

Yes, we have to be really careful with file uploads because if the user is running their script a million times they could upload a million files to our S3 bucket, and since we rename each file with UUID, it would never even recognize that this file has already been uploaded and will just upload it again every time. We could get stuck paying a hefty bill for a bunch of test files uploaded to our S3.

I think your solution would delay this process, but wouldn't solve it completely. The most improvement to this would be if we place the file upload at the very last part of save, so all checks would pass and then it would upload which would help catch errors and reduce bad file uploads, but doesn't solve it completely. 

I think we may want to add a flag to the file node with something like `local_file_upload=False` and if it is false, then we don't upload and just return, otherwise we upload. This could be used for debugging purposes when the user is first writing their script originally.

src/cript/api/api.py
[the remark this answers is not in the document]
InnocentBug: agreed. I changed it to a single underscore and a name that indicates internal usage only.

src/cript/api/api.py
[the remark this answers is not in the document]
InnocentBug: I don't know. This is just temporary anyways.
So we should just use something that works. and make it note to do it right, once we can.

nh916:
> No, I need this now.
I am in the situation that I am working and debugging CRIPT scripts.
So the aforementioned situation.
I don't wait for uploading files all the time, plus I don't want to fill the S3 bucket with so many trash files.

what about the idea of adding an `upload_local_file=False` idea? Would that solve the majority of the problem?

src/cript/api/api.py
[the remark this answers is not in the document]
nh916: okay that makes sense

nh916:
I really like the idea of having all checks first pass before we start uploading files because then we add a layer of protection to our code so we don't upload any test data.

InnocentBug:
> I really like the idea of having all checks first pass before we start uploading files because then we add a layer of protection to our code so we don't upload any test data.

which tests are you talking about?

In this design, `project.validate` passed before we call upload.
Are there any other tests we can perform locally?

Yes, Ideally we'd upload the files after the nodes are accepted by the back end.
But that either requires us to upload files with not yet existing links into the back end (and fill the links later with uploads.)
Or we need to patch the files nodes after the file upload, to represent the new uploaded file name.

Since we don't have PATCH yet, and I don't like the first option.

This temproray fix is the best I can come up with.

nh916:
> > I really like the idea of having all checks first pass before we start uploading files because then we add a layer of protection to our code so we don't upload any test data.
> 
> which tests are you talking about?
> 
> In this design, `project.validate` passed before we call upload. Are there any other tests we can perform locally?
> 
> Yes, Ideally we'd upload the files after the nodes are accepted by the back end. But that either requires us to upload files with not yet existing links into the back end (and fill the links later with uploads.) Or we need to patch the files nodes after the file upload, to represent the new uploaded file name.
> 
> Since we don't have PATCH yet, and I don't like the first option.
> 
> This temproray fix is the best I can come up with.

no I was just saying that I like your idea of having a layer of protection for file uploads and only allowing uploads after db schema and any other checks that we have for the user has passed, that way there are less mistaken uploads

src/cript/api/api.py
nh916, line 197:
```suggestion
aws_token = aws_token.lstrip("Bearer ")
```

nh916: please consider changing this one this is easy and will make the code cleaner, I know that it's temporary but lets make our code clean anyways

src/cript/nodes/supporting_nodes/file.py
```diff
@@ -152,25 +153,50 @@ def __init__(self, name: str, source: str, type: str, extension: str = "", data_
...
-            source=source,
+            # always giving the function the required str regardless if the input `Path` or `str`
+            source=str(source),
             extension=extension,
             data_dictionary=data_dictionary,
         )
 
         self.validate()
 
+    def ensure_uploaded(self, api=None):
```
nh916, line 171: does this need to be a public method or can it be private?

nh916: can we remove passing of the API and just have it use the global API?

src/cript/api/api.py
```diff
@@ -513,6 +521,11 @@ def save(self, project: Project) -> None:
         None
             Just sends a `POST` or `Patch` request to the API
         """
+
+        # Ensure that all file nodes have uploaded there payload before actual save.
+        for file_node in project.find_children({"node": ["File"]}):
+            file_node.ensure_uploaded(api=self)
```
nh916, line 527: I really dislike this part of the design because it seems confusing to me, but I can accept it for a temporary solution and later we think of something better

nh916 (changes requested):

src/cript/api/api.py
```diff
@@ -207,6 +208,13 @@ def _get_s3_client(self) -> boto3.client:  # type: ignore
 
         return s3_client
 
+    # Use a property to ensure delayed init of s3_client
+    @property
+    def _s3_client(self):
+        if self._internal_s3_client is None:
+            self._internal_s3_client = self._get_s3_client()
```
nh916, line 215: I just noticed that the creation of the S3 client is split up along 2 different methods. I would recommend against this because it creates more confusion for the users and developers because they have to figure out which method does what when they are both essentially have the same purpose. I would just put all the code for the creation of the s3 client within a single method and be done with it

src/cript/api/api.py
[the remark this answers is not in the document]
InnocentBug: I add what, I can, but it isn't great.

src/cript/api/api.py
InnocentBug: ok, this is implementation details of find_children.
We can later reimplement it as we find better solutions.

(The best that comes to mind, is that we have a variety of caches (uuid hash based), that let us look up stuff.
For example: we have a cache of all "file" nodes, and we keep a look-up map for all nodes in a project.
But both of those ideas require additional code, making the SDK less lightweight.)

I am open to add something like this, later so far we have pushed stuff like this back.

src/cript/nodes/supporting_nodes/file.py
InnocentBug: I added a doc string.

InnocentBug:
> what about the idea of adding an `upload_local_file=False` idea? Would that solve the majority of the problem?

I can add this, but I am not convinced it is great.
Users might forget to change this after they are done debugging.
(Or set it as a matter of "fixing" a problem with their file upload.)
And in that scenario, we get a bunch of "useless" file nodes in the CRIPT databank.
I am actually no sure, what for example the front end would do if the file node only has a local file name?

InnocentBug:
> > 
> 
> no I was just saying that I like your idea of having a layer of protection for file uploads and only allowing uploads after db schema and any other checks that we have for the user has passed, that way there are less mistaken uploads

Ok, we are on the same page here.
I add a line, that calls `project.validate` before the upload loop is started.
So, we perform all the local tests we can first.

(It adds another validate call to project though, because `.json` call also calls `validate`. But maybe that is something for optimization (i.e. keep a cache with hashed valid nodes.))

src/cript/api/api.py
InnocentBug: sure, I add it.

src/cript/nodes/supporting_nodes/file.py
InnocentBug: Yes, if a user is aware of this functionality, I wouldn't want to limit them from optimization.
So, making it public allows the user to take control, when it is beneficial.
I wouldn't want to limit able users.
(The doc string states, that it will be called automatically.)

InnocentBug: I would like to leave the API in there. Exactly for the public usage.
It appears more transparent to the user if they can choose the API.

And it gives them more options. For the auto upload and the uninformed user, it exactly looks as before.

Also, it just makes sense, that if you call `api.save(project)` that the `api` object is used to perform the upload.
So if we don't rely on the global cache option there, we shouldn't rely on it for the upload either.

src/cript/api/api.py
InnocentBug: If you have a better one, let me know.

(One thing, I can think of, that we integrate the `ensure_upload` call into the JSON serialization.
This requires no extra searching of file nodes, they just get uploaded while converted to JSON.
However, this ALWAYS happens BEFORE DB_Schema validation, since we validate the JSON. So we can't perform validation before upload. I don't think that would be great.)

src/cript/api/api.py
InnocentBug: I thought it was clearer, to keep the auth and contruction out of the caching mechanism.
But if you prefer it in a single function.
No problem. I combined them.

src/cript/api/api.py
nh916: oof you are right!

The serialization is a good idea because it makes logical sense, but then we get into the issue that you pointed out earlier that we might upload a ton of files during debugging.

okay I think this should work for now. I think API class needs refactoring anyways because we outgrew our original design, and we can tackle it all there.

InnocentBug: One thought may be to rewrite the serialization, to conditionally disable the upload there.
(And only activate the upload when the JSON is generated for `save`).

I could implement that.
If you prefer that, I can do that.
Probably takes only a few hours tops.

nh916:
> Users might forget to change this after they are done debugging.

we could put out a warning in the console if the user is in debug mode?

```bash
Debug Warning: You are running the CRIPT Python SDK in debug mode. Please turn off debug mode before saving to CRIPT. For detailed look please visit the documentation.
```

and then in the documentation we can have a page that says what will be disabled, how to use it, when to use it, and how to change it.

I think the con of this is that it might introduce more errors because now we might have to catch the errors in case of debug mode.

could be a cool idea, but lets come back to it later

nh916 (approved):

InnocentBug:
> > Users might forget to change this after they are done debugging.
> 
> we could put out a warning in the console if the user is in debug mode?
> 
> ```shell
> Debug Warning: You are running the CRIPT Python SDK in debug mode. 
> Please turn off debug mode before saving to CRIPT. For detailed look please visit the documentation.
> ```
> 
> and then in the documentation we can have a page that says what will be disabled, how to use it, when to use it, and how to change it.
> 
> I think the con of this is that it might introduce more errors because now we might have to catch the errors in case of debug mode.
> 
> could be a cool idea, but lets come back to it later for V2

Yes, that could be a cool idea.
But my feeling is, that most people wouldn't use it, since they don't have an upside for using debug mode.

nh916:
> > > Users might forget to change this after they are done debugging.
> > 
> > 
> > we could put out a warning in the console if the user is in debug mode?
> > ```shell
> > Debug Warning: You are running the CRIPT Python SDK in debug mode. 
> > Please turn off debug mode before saving to CRIPT. For detailed look please visit the documentation.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then in the documentation we can have a page that says what will be disabled, how to use it, when to use it, and how to change it.
> > I think the con of this is that it might introduce more errors because now we might have to catch the errors in case of debug mode.
> > could be a cool idea, but lets come back to it later for V2
> 
> Yes, that could be a cool idea. But my feeling is, that most people wouldn't use it, since they don't have an upside for using debug mode.

that is a good point!

PR #183, opened 2023-06-27
Keep track of already validated nodes

InnocentBug:
# Description
This may lead to a smaller performance boost.

This keeps track of already validated node, by storing a simplified hash value of already validated JSON attributes.

## Changes

Added a `sloppy` argument to `validate` that can disable the hash based accelerated validation.
Added additional `validate` before projects are saved.

## Tests

our usual tests, should cover this functionality.
I have not performed explicit testing to measure the acceleration.
The possible acceleration depends on the specific situation, especially validating large graphs multiple times, can be accelerated this way. 

## Known Issues

The hash value is intentionally set up, to not include a full graph traversal.
Breaking changes beyond the immediate children are possibly not detected.

For project, a special treatment is used, where the full graph is traversed, since project nodes have to be fully traversed to detect orphan nodes.
## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -513,6 +513,10 @@ class API:
         None
             Just sends a `POST` or `Patch` request to the API
         """
+
+        # Ensure, that the project is valid with this API.
+        project.validate(api=self, sloppy=False)
+
         response: Dict = requests.post(url=f"{self._host}/{project.node_type.lower()}", headers=self._http_headers, data=project.json).json()
 
         # if http response is not 200 then show the API error to the user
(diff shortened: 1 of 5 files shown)


nh916:
I am not sure yet how much of an optimization this is so I am not sure how urgently I should treat this. For speed optimizations, I really like [cProfile in Python](https://docs.python.org/3/library/profile.html) that is really good at giving a good break down, and I think we can run one of our bigger integration tests to see the difference in speed because those are bigger.

Currently, I think we are too close and I want to keep changes to the minimum and just focus on stabilizing the current code that we have.

I'm thinking since we are going to do more optimizations later anyways, lets push this one for later

InnocentBug:
Close for now, Idea still valid.

PR #186, opened 2023-07-06
Bad UUID handling

InnocentBug:
# Description

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -548,7 +548,7 @@ class API:
         if node_known:
             response: Dict = requests.patch(url=f"{self._host}/{node.node_type.lower()}", headers=self._http_headers, data=json_data).json()
         else:
-            response: Dict = requests.post(url=f"{self._host}/{node.node_type.lower()}", headers=self._http_headers, data=json_data).json()
+            response: Dict = requests.post(url=f"{self._host}/{node.node_type.lower()}", headers=self._http_headers, data=json_data).json()  # type: ignore
 
         # If we get a Bad UUID we need to handle this extra and save the Bad node first.
         if response["code"] == 400 and response["error"].startswith("Bad uuid:"):
@@ -563,7 +563,7 @@ class API:
 
             # Now we save the bad node extra.
             # So it will be known when we attempt to save the graph again.
-            known_uuid.union(self._internal_save(missing_node, known_uuid))
+            known_uuid.union(self._internal_save(missing_node, known_uuid))  # type: ignore
             # Since we pre-saved this node, we want it to be UUID edge only the next JSON.
             # So we add it to the list of known nodes
             known_uuid.add(missing_uuid)
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -491,7 +491,7 @@ class API:
         # if validation goes through without any problems return True
         return True
 
-    def save(self, node, known_uuid: Optional[Set[str]] = None) -> Optional[Set[str]]:
+    def save(self, project: Project) -> None:
         """
         This method takes a project node, serializes the class into JSON
         and then sends the JSON to be saved to the API.
@@ -513,6 +513,24 @@ class API:
         A set of extra saved node UUIDs.
             Just sends a `POST` or `Patch` request to the API
         """
+        try:
+            prehandled_uuid = self._internal_save(project)
+        except Exception as exc:
+            # TODO remove all prehandled nodes.
+            raise exc
+
+    def _internal_save(self, node, known_uuid: Optional[Set[str]] = None) -> Optional[Set[str]]:
+        """
+        Internal helper function that handles the saving of different nodes (not just project).
+
+        If a "Bad UUID" error happens, we find that node with the UUID and save it first.
+        Then we recursively call the _internal_save again.
+        Because it is recursive, this repeats until no "Bad UUID" error happen anymore.
+        This works, because we keep track of "Bad UUID" handled nodes, and represent them in the JSON only as the UUID.
+        """
+
+        # known_uuid are node, that we have saved to the back end before.
+        # We keep track of it, so that we can condense them to UUID only in the JSON.
         if known_uuid is None:
             known_uuid = set()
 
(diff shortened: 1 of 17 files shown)


src/cript/api/paginator.py
```diff
@@ -193,11 +193,17 @@ def fetch_page_from_api(self) -> List[dict]:
...
-        if "result" in response["data"]:
+        try:
             self.current_page_results = response["data"]["result"]
-        else:
+        except KeyError:
+            self.current_page_results = response["data"]
+        except TypeError:
             self.current_page_results = response["data"]
 
+        if response["code"] == 404 and response["error"] == "The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.":
```
nh916, line 203: do you want to put the paginator refactoring in a different PR?

nh916: has this been tested with the search test functions?

src/cript/api/api.py
```diff
@@ -511,22 +510,42 @@ def save(self, project: Project) -> None:
...
-        None
+        A set of extra saved node UUIDs.
             Just sends a `POST` or `Patch` request to the API
         """
+        if known_uuid is None:
+            known_uuid = set()
 
-        project.validate()
-
+        node.validate()
```
nh916, line 519: I see we swapped project to node. I had a similar idea where I still have the same save for the project only, but there is a helper method that takes just a node and saves that to the API, so we can reuse the code within our brute force save method

src/cript/api/api.py
```diff
@@ -511,22 +510,42 @@ def save(self, project: Project) -> None:
...
         # saves all the local files to cloud storage right before saving the Project node
         # Ensure that all file nodes have uploaded there payload before actual save.
-        for file_node in project.find_children({"node": ["File"]}):
+        for file_node in node.find_children({"node": ["File"]}):
```
nh916, line 522: I think if we `use a send_post_request(node)` helper method then we can leave the current code as is and the user only thinks they are saving a project still

src/cript/api/api.py
```diff
@@ -511,22 +510,42 @@ def save(self, project: Project) -> None:
...
             file_node.ensure_uploaded(api=self)
 
-        response: Dict = requests.post(url=f"{self._host}/{project.node_type.lower()}", headers=self._http_headers, data=project.json).json()
+        node_known = len(self.search(type(node), SearchModes.UUID, str(node.uuid)).current_page_results) == 1
```
nh916, line 525: I wrote a `get_bad_uuid_from_error_message` that I think would be helpful for getting the UUID from the error message

src/cript/api/api.py
```diff
@@ -511,22 +510,42 @@ def save(self, project: Project) -> None:
...
 
+        json_data = node.get_json(known_uuid=known_uuid).json
+        if node_known:
+            response: Dict = requests.patch(url=f"{self._host}/{node.node_type.lower()}", headers=self._http_headers, data=json_data).json()
```
nh916, line 529: I am guessing we'll have to send it in a specific order because otherwise we can send the project as patch and it would give us issues

src/cript/api/api.py
```diff
@@ -511,22 +510,42 @@ def save(self, project: Project) -> None:
...
+        else:
+            response: Dict = requests.post(url=f"{self._host}/{node.node_type.lower()}", headers=self._http_headers, data=json_data).json()
+
+        if response["code"] == 400 and response["error"].startswith("Bad uuid:"):
```
nh916, line 533: I catch this within the `brute_force_save` function and handle it there. I'd suggest using that if possible

src/cript/api/api.py
```diff
@@ -511,22 +510,42 @@ def save(self, project: Project) -> None:
...
+            # We can have multiple occurrences of the node,
+            # but it doesn't matter which one we save
+            # TODO some error handling, for the BUG case of not finding the UUID
+            missing_node = node.find_children({"uuid": missing_uuid})[0]
+            known_uuid = self.save(missing_node, known_uuid)
+            known_uuid.add(missing_uuid)
+            # Recursive call to finish the post
+            return self.save(node, known_uuid)
+
+        print("asdf", response)
```
nh916, line 544: does this need to be removed?

src/cript/nodes/core.py
```diff
@@ -260,6 +261,9 @@ class ReturnTuple:
         previous_handled_nodes = copy.deepcopy(NodeEncoder.handled_ids)
         if handled_ids is not None:
             NodeEncoder.handled_ids = handled_ids
+        previous_known_uuid = copy.deepcopy(NodeEncoder.known_uuid)
```
nh916, line 264: I advise against deepcopy whenever possible just because I had a very hard time when working with fixtures that used it for the integration tests. It just made things more complicated, but if we have to use it then we can use it

src/cript/nodes/util/__init__.py
```diff
@@ -37,6 +38,14 @@ def default(self, obj):
                 if uid in NodeEncoder.handled_ids:
                     return {"node": obj._json_attrs.node, "uid": uid}
 
+            try:
+                uuid_str = str(obj.uuid)
```
nh916, line 42: obj is a bad variable name, can we name it something more descriptive?

src/cript/nodes/util/__init__.py
```diff
@@ -37,6 +38,14 @@ def default(self, obj):
...
+            except AttributeError:
```
nh916, line 43: why are we catching an error and not doing anything with it?

nh916:
honestly, I have a bit of a hard time understanding the full code

src/cript/api/paginator.py
InnocentBug: Well, the paginator was raising a cryptic exception when searching for a non-existent node.

But for the patch solution, we need it to return an empty list in that case.

I put it in a different commit, but I need this here to work.

src/cript/api/api.py
InnocentBug: This is just a name here.

It doesn't change anything for the user.

It is just for devs. For a dev it would be confusing to call it project, since it can be other things, when called differently.

I am ok, wrapping that a little differently, such that we expose one with `project` to the user, but we can do it after we confirm this works.

src/cript/api/api.py
InnocentBug: see comment above.

src/cript/api/api.py
InnocentBug: yeah, I didn't see or use that function. we can use it here.

InnocentBug: well, actually no.

This is doing something else then you think.

This just checks if a UUID is known in the backend.
We can use a helper function for that too, if you like.

src/cript/api/api.py
InnocentBug: No, that shouldn't be necessary.
The only thing that I haven't handled yet, are `duplicate node` errors.
Everything else should automatically put itself in the right order.

src/cript/api/paginator.py
nh916: got it!

We have to come back to paginator later and really beef it up and refactor it because when I first wrote it it was just a rough draft of what I knew then

src/cript/api/api.py
InnocentBug: we can, use a helper function to catch this.
But your `brute_force_save` function, works fundamentally different then my save.

src/cript/api/api.py
InnocentBug: yeah, forget to remove.

src/cript/nodes/core.py
InnocentBug: No, there is a deepcopy necessary here.

However, this is a deepcopy of a list of strings. Not of nodes.
So very different then your issues in the fixtures.

The fixture issues are only caused, because DB schema doesn't allow UID where it should sometimes. We need to fix that separately.

src/cript/api/api.py
nh916: I agree, if we give the user to save only with project and we have a hidden one that works with node, that would be better. Yeah I agree we can do it after this works

src/cript/nodes/util/__init__.py
InnocentBug: This is the variable name that the documentation of the `json` module uses.
We should not deviate from that.

src/cript/nodes/util/__init__.py
InnocentBug: yes, that is a very pythonic way.

Instead of testing if an attribute is present, this is preferred.

src/cript/api/api.py
nh916: I do like helper functions!

nh916: but I agree with you, lets get it working and then lets refactor it to make it better

src/cript/api/api.py
nh916: okie dokes, that sounds good. It feels too simple though

src/cript/api/api.py
nh916: okie dokes we can get something working for now and come back and refactor it after

src/cript/nodes/core.py
nh916: okie dokes, that sounds good!

src/cript/nodes/util/__init__.py
nh916: I agree, the principle of `try instead of checking for if`

src/cript/nodes/util/__init__.py
nh916: okie dokes, that makes sense

nh916:
@InnocentBug made a small commit to fix the merge conflict issue

nh916:
@InnocentBug I see a few errors on our side but mostly on API side, do we want to fix them before merging it in, or do you think this is good enough to approve for now and we can fix the rest later?

[Python-SDK-Test-Results-html-2023-07-06-17_06_26.pdf](https://github.com/C-Accel-CRIPT/Python-SDK/files/11973356/Python-SDK-Test-Results-html-2023-07-06-17_06_26.pdf)

InnocentBug:
Yeah, there are a couple of errors left. 
I am working on fixing some of the errors but we can merge it before all errors are fixed.

nh916 (approved):

PR #189, opened 2023-07-07
add a node cache to the UUIDBaseNode. This node cache is used to upda…

InnocentBug:
…te existing UUID nodes, rather then creating a new node with the same UUID.

# Description

## Checklist

### DIFF
--- a/src/cript/nodes/core.py
+++ b/src/cript/nodes/core.py
@@ -166,13 +166,20 @@ class BaseNode(ABC):
             else:
                 arguments[field] = json_dict[field]
 
-        try:
-            node = cls(**arguments)
-        # TODO we should not catch all exceptions if we are handling them, and instead let it fail
-        #  to create a good error message that points to the correct place that it failed to make debugging easier
-        except Exception as exc:
-            print(cls, arguments)
-            raise exc
+        # If a node with this UUID already exists, we don't create a new node.
+        # Instead we use the existing node from the cache and just update it.
+        from cript.nodes.uuid_base import UUIDBaseNode
+
+        if "uuid" in json_dict and json_dict["uuid"] in UUIDBaseNode._uuid_cache:
+            node = UUIDBaseNode._uuid_cache[json_dict["uuid"]]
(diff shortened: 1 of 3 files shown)


src/cript/nodes/core.py
```diff
@@ -166,13 +166,20 @@ def _from_json(cls, json_dict: dict):
...
+
+        if "uuid" in json_dict and json_dict["uuid"] in UUIDBaseNode._uuid_cache:
+            node = UUIDBaseNode._uuid_cache[json_dict["uuid"]]
+        else:  # Create a new node
+            try:
+                node = cls(**arguments)
+            # TODO we should not catch all exceptions if we are handling them, and instead let it fail
+            #  to create a good error message that points to the correct place that it failed to make debugging easier
+            except Exception as exc:
+                print(cls, arguments)
```
nh916, line 181: does this need to be removed?

src/cript/nodes/core.py
```diff
@@ -166,13 +166,20 @@ def _from_json(cls, json_dict: dict):
...
-            print(cls, arguments)
-            raise exc
+        # If a node with this UUID already exists, we don't create a new node.
+        # Instead we use the existing node from the cache and just update it.
+        from cript.nodes.uuid_base import UUIDBaseNode
+
+        if "uuid" in json_dict and json_dict["uuid"] in UUIDBaseNode._uuid_cache:
+            node = UUIDBaseNode._uuid_cache[json_dict["uuid"]]
+        else:  # Create a new node
+            try:
```
nh916, line 176: I see you check for cache hit, but in the case of cache miss are you adding it to the cache?

InnocentBug: that happens automatically in the UUIDBaseNode constructor

src/cript/nodes/core.py
InnocentBug: maybe, we had it in there even before I changed it

nh916 (approved):

PR #190, opened 2023-07-07
Refactor the save a little bit. Patch does not work.

InnocentBug:
# Description

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -21,6 +21,7 @@ from cript.api.exceptions import (
 )
 from cript.api.paginator import Paginator
 from cript.api.utils.get_host_token import resolve_host_and_token
+from cript.api.utils.save_helper import fix_node_save
 from cript.api.valid_search_modes import SearchModes
 from cript.api.vocabulary_categories import ControlledVocabularyCategories
 from cript.nodes.exceptions import CRIPTJsonNodeError, CRIPTNodeSchemaError
@@ -515,7 +516,7 @@ class API:
             self._internal_save(project)
         except Exception as exc:
             # TODO remove all pre-handled nodes.
-            raise exc
+            raise exc from exc
 
     def _internal_save(self, node, known_uuid: Optional[Set[str]] = None) -> Optional[Set[str]]:
         """
@@ -546,30 +547,19 @@ class API:
         # if it does exist we use `patch` if it doesn't `post`.
         node_known = len(self.search(type(node), SearchModes.UUID, str(node.uuid)).current_page_results) == 1
         if node_known:
-            response: Dict = requests.patch(url=f"{self._host}/{node.node_type.lower()}", headers=self._http_headers, data=json_data).json()
+            response: Dict = requests.patch(url=f"{self._host}/{node.node_type.lower()}/{str(node.uuid)}", headers=self._http_headers, data=json_data).json()
         else:
             response: Dict = requests.post(url=f"{self._host}/{node.node_type.lower()}", headers=self._http_headers, data=json_data).json()  # type: ignore
 
-        # If we get a Bad UUID we need to handle this extra and save the Bad node first.
-        if response["code"] == 400 and response["error"].startswith("Bad uuid:"):
-            # We extract the
-            missing_uuid = response["error"][len("Bad uuid: ") : -len(" provided")]
+        # If we get an error we may be able to fix, we to handle this extra and save the bad node first.
+        # Errors with this code, may be fixable
+        if response["code"] in (400, 409):
+            nodes_fixed = fix_node_save(self, node, response, known_uuid)
+            # In case of a success, we return the know uuid
+            if nodes_fixed is not False:
+                return nodes_fixed
+            # if not successful, we escalate the problem further
 
-            # Use the find_children functionality to find that node in our current tree
-            # We can have multiple occurrences of the node,
-            # but it doesn't matter which one we save
-            # TODO some error handling, for the BUG case of not finding the UUID
-            missing_node = node.find_children({"uuid": missing_uuid})[0]
-
-            # Now we save the bad node extra.
-            # So it will be known when we attempt to save the graph again.
-            known_uuid.union(self._internal_save(missing_node, known_uuid))  # type: ignore
-            # Since we pre-saved this node, we want it to be UUID edge only the next JSON.
-            # So we add it to the list of known nodes
-            known_uuid.add(missing_uuid)
-            # Recursive call.
-            # Since we should have fixed the "Bad UUID" now, we can try to save the node again
-            return self._internal_save(node, known_uuid)
         if response["code"] != 200:
             raise CRIPTAPISaveError(api_host_domain=self._host, http_code=response["code"], api_response=response["error"])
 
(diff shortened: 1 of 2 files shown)


src/cript/api/utils/save_helper.py
```diff
@@ -1,50 +1,29 @@
...
-
-            # there was an error with the save, so loop over again and try again
-
-            # if we are not handling the error then raise it
-            else:
-                raise error
-
-
-def get_bad_uuid_from_error_message(error_message: str) -> str:
+    assert response["code"] in (400, 409)
```
nh916, line 8: I suggest using assert within only tests, its a bit strange seeing it in code, but if we want to keep it I think that should work too

src/cript/api/api.py
```diff
@@ -546,30 +547,19 @@ def _internal_save(self, node, known_uuid: Optional[Set[str]] = None) -> Optiona
...
-            known_uuid.union(self._internal_save(missing_node, known_uuid))  # type: ignore
-            # Since we pre-saved this node, we want it to be UUID edge only the next JSON.
-            # So we add it to the list of known nodes
-            known_uuid.add(missing_uuid)
-            # Recursive call.
-            # Since we should have fixed the "Bad UUID" now, we can try to save the node again
-            return self._internal_save(node, known_uuid)
+        # If we get an error we may be able to fix, we to handle this extra and save the bad node first.
+        # Errors with this code, may be fixable
+        if response["code"] in (400, 409):
```
nh916, line 556: I suggest having a variable for the list of errors that we handle, in case we want to add or subtract to it we can do so later

```suggestion

http_status_code_handled = (400, 409)

# If we get an error we may be able to fix, we to handle this extra and save the bad node first.
# Errors with this code, may be fixable
if response["code"] in http_status_code_handled
```

nh916: I don't know if we want to rely on the status codes because `HTTP 400` just means generic bad request and can be used for many things. I think this is good for now but at some point we'll need to have solid errors that we handle and we can rely on. Those errors have to for sure be variables for us to be able to change easily if the backend changes

src/cript/api/utils/save_helper.py
```diff
@@ -1,50 +1,29 @@
...
+    if response["error"].startswith("Bad uuid:") or response["error"].strip().startswith("Duplicate uuid:"):
```
nh916, line 9: can we put this into a function and just feed it the error and then within that function we have a kind of switch statement that checks the error against a bunch of cases that we handle, at the end if it doesn't match anything then we raise the API error

src/cript/api/utils/save_helper.py
InnocentBug: I changed this part now, I raise an exception instead of assert.

Yes, assert should not really be in production code (although it doesn't hurt either).

src/cript/api/utils/save_helper.py
InnocentBug: That is already what we are doing here.

If the implementation can't handle the error, it will just pass it through and the outer `_internal_save` function handles the API error.

InnocentBug: Notice, how I return `False` if that if block isn't hit right.
That is the indication of the outer function, that the error has not been handled.

src/cript/api/api.py
nh916: okay that is fair

nh916: I would still suggest to put the error codes in a variable, but I can be okay with it if we decide not to do that for now

```python
http_status_code_handled = (400, 409)
# If we get an error we may be able to fix, we to handle this extra and save the bad node first.
# Errors with this code, may be fixable
if response["code"] in http_status_code_handled
```

src/cript/api/utils/save_helper.py
nh916: from what I understand the function either returns another function or returns False, I don't love that because I think it is best if a function only returns a single thing, would it make sense if the function either calls another function or throws an error or returns None or something.

src/cript/api/utils/save_helper.py
```diff
@@ -1,50 +1,29 @@
...
+        # Now we save the bad node extra.
+        # So it will be known when we attempt to save the graph again.
+        # Since we pre-saved this node, we want it to be UUID edge only the next JSON.
+        # So we add it to the list of known nodes
+        known_uuid.union(api._internal_save(missing_node, known_uuid))  # type: ignore
+        # The missing node, is now known to the API
+        known_uuid.add(missing_uuid)
+        # Recursive call.
+        # Since we should have fixed the "Bad UUID" now, we can try to save the node again
+        return api._internal_save(node, known_uuid)
```
nh916, line 22: this line scares me a bit, if this gets repeated 1,000 times, can it hit stack overflow error?

src/cript/api/utils/save_helper.py
InnocentBug: No, it returns `False` or a set of `known_uuid`s.

That seems fairly straight forward to me.
You need the `known_uuid` for the recursion.

And False, to signal the error was not processed.

src/cript/api/utils/save_helper.py
InnocentBug: it definitely could.
It is just not likely to happen.

That would mean 1,000 errors from the API.

We could technically refactor it to not use recursion, but that doesn't seem simple or straight forward right now.

nh916: okie dokes, I think that works for now

src/cript/api/api.py
nh916: I see your logic, usually I make it into variables because it makes it easier to update in the future, but I think it is fine for now and we can update it later

nh916:
should we wait on https://github.com/C-Accel-CRIPT/Python-SDK/pull/193 to be merged first before approving this one?

InnocentBug:
> should we wait on #193 to be merged first before approving this one?

no, I don't think we need to wait.
#193 needs some more work

src/cript/api/utils/save_helper.py
```diff
@@ -1,50 +1,29 @@
...
+        missing_uuid = get_uuid_from_error_message(response["error"])
+        missing_node = find_node_by_uuid(node, missing_uuid)
+
+        # Now we save the bad node extra.
+        # So it will be known when we attempt to save the graph again.
+        # Since we pre-saved this node, we want it to be UUID edge only the next JSON.
+        # So we add it to the list of known nodes
+        known_uuid.union(api._internal_save(missing_node, known_uuid))  # type: ignore
+        # The missing node, is now known to the API
+        known_uuid.add(missing_uuid)
```
nh916, line 19: I think `known_uuid` might be better named `saved_nodes_uuid`, but a name doesn't make much of a difference

src/cript/api/utils/save_helper.py
```diff
@@ -58,26 +37,19 @@ def get_bad_uuid_from_error_message(error_message: str) -> str:
...
 
-    # TODO This should not be triggered and should be handled better
-    raise Exception("could not find any material with that UUID")
 
+def find_node_by_uuid(node, uuid: str):
+    # Use the find_children functionality to find that node in our current tree
+    # We can have multiple occurrences of the node,
+    # but it doesn't matter which one we save
+    # TODO some error handling, for the BUG case of not finding the UUID
+    missing_node = node.find_children({"uuid": uuid})[0]
```
nh916, line 53: nicely done!

nh916 (approved):

PR #192, opened 2023-07-08
added JSON representation to all nodes in documentation

nh916:
# Description
Added JSON representation of every node within its documentation page. I think this helps a lot to understand what the code under the hood is doing and helps the user understand and work with the software better.

## Changes
* Add a block of JSON representation code to every node documentation page.
* formatted the process node table
* fixed broken links within data.py
    * formatted the table as well

## Known Issues
* The JSON has to be updated manually every time something changes
* Some of the JSON can be better picked to easier demonstrate how the code works to the user
    * However, what it has currently I think helps a lot too

## Screenshots

<details>
<summary>click to view the screenshots</summary>

image
image
image
</details>

### DIFF
--- a/src/cript/nodes/primary_nodes/inventory.py
+++ b/src/cript/nodes/primary_nodes/inventory.py
@@ -37,7 +37,7 @@ class Inventory(PrimaryBaseNode):
              "name":"my material 1",
              "uid":"_:9679ff12-f9b4-41f4-be95-080b78fa71fd",
              "uuid":"9679ff12-f9b4-41f4-be95-080b78fa71fd"
-             "bigsmiles":"CC(C)C1=CC=CC=C1",
+             "bigsmiles":"[H]{[>][<]C(C[>])c1ccccc1[]}",
           },
           {
              "node":["Material"],
--- a/src/cript/nodes/primary_nodes/inventory.py
+++ b/src/cript/nodes/primary_nodes/inventory.py
@@ -24,6 +24,32 @@ class Inventory(PrimaryBaseNode):
     | material  | list[[Material](./material.md)] |                     | material that you like to group together |
 
 
+    ## JSON Representation
+    ```json
+    {
+       "name":"my inventory name",
+       "node":["Inventory"],
+       "uid":"_:90f45778-b7c9-4b77-8b83-a6ea9671a937",
+       "uuid":"90f45778-b7c9-4b77-8b83-a6ea9671a937",
+       "material":[
+          {
+             "node":["Material"],
+             "name":"my material 1",
+             "uid":"_:9679ff12-f9b4-41f4-be95-080b78fa71fd",
+             "uuid":"9679ff12-f9b4-41f4-be95-080b78fa71fd"
+             "bigsmiles":"123456",
+          },
+          {
+             "node":["Material"],
+             "name":"my material 2",
+             "uid":"_:1ee41708-3531-43eb-8049-4bb91ad73df6",
+             "uuid":"1ee41708-3531-43eb-8049-4bb91ad73df6"
+             "bigsmiles":"654321",
+          }
+       ]
+    }
+    ```
+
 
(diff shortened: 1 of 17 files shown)


src/cript/nodes/primary_nodes/inventory.py
```diff
@@ -24,6 +24,32 @@ class Inventory(PrimaryBaseNode):
...
+    ## JSON Representation
+    ```json
+    {
+       "name":"my inventory name",
+       "node":["Inventory"],
+       "uid":"_:90f45778-b7c9-4b77-8b83-a6ea9671a937",
+       "uuid":"90f45778-b7c9-4b77-8b83-a6ea9671a937",
+       "material":[
+          {
+             "node":["Material"],
```
InnocentBug, line 36: That is actually not how it would come out if you called `.json`.
The material has to be condensed into UUID to be valid under the schema.

src/cript/nodes/primary_nodes/material.py
```diff
@@ -46,21 +46,13 @@ class Material(PrimaryBaseNode):
...
-        "uid": "0x24a08",
-        "updated_at": "2023-03-14T00:45:02.196276Z",
-        "uuid": "403fa02c-9a84-4f9e-903c-35e535151b08",
-    }
+      {
+         "node":["Material"],
+         "name":"my unique material name",
+         "uid":"_:9679ff12-f9b4-41f4-be95-080b78fa71fd",
+         "uuid":"9679ff12-f9b4-41f4-be95-080b78fa71fd"
+         "bigsmiles":"123456",
```
InnocentBug, line 54: for an example, we should use a valid bigsmiles

InnocentBug (approved):

src/cript/nodes/primary_nodes/inventory.py
nh916: I was thinking of not putting the UUID for the user for now because it might confuse them, but we could have that JSON and write a small section on condensing the node to its UUID and link to it in all places that it is condensed.

src/cript/nodes/primary_nodes/material.py
nh916: what would I put for this material? I could put the bigsmiles for polystyrene, but not sure if that makes sense or if there is something better.

nh916: also feel free to make updates to this branch as you see fit, I just put something there for now to build on as we go

InnocentBug: yeah, polystyrene should be good enough.

We just shouldn't advertise, that we allow invalid bigsmiles strings here.

src/cript/nodes/primary_nodes/inventory.py
InnocentBug: yeah, ok that is fair.

InnocentBug (approved):

src/cript/nodes/primary_nodes/material.py
nh916: that is very true, I'll update it now!

InnocentBug (approved):
That actually isn't poly (styrene) that is the SMILES string for styrene.

But since SMILES is a subset of BigSMILES, that should be fine.
This is BigSMILES for poly(styrene) `[H]{[>][<]C(C[>])c1ccccc1[]}`

I also look it up every time, just to be sure.

nh916:
> That actually isn't poly (styrene) that is the SMILES string for styrene.
> 
> But since SMILES is a subset of BigSMILES, that should be fine. This is BigSMILES for poly(styrene) `[H]{[>][<]C(C[>])c1ccccc1[]}`
> 
> I also look it up every time, just to be sure.

done! ✅ 

let me know if you find any other upgrades and I can add them pretty quickly

InnocentBug (approved):

PR #193, opened 2023-07-10
Handle patch error messages with SDK

InnocentBug:
# Description
We read the error message from the back end in patches.
And we analyze them to adjust our JSON to be compliant.

Similar strategy as with `Bad UUID`.

## Changes

Minor changes to the save routine to catch and handle those errors.

The serialization now optionally allows the suppression attributes, to be compliant with expectation of back end.

This also refactors the internal save function a bit, making it clearer, of what needs to be omitted from JSON for the actual save.

## Tests
None, yet.
We need to come up with some.

## Known Issues

Only implemented for fields on the `/` path.
I haven't seen another response, so if we have an example for deeper modification, we can/have to implement it.

## Checklist


nh916:
I can write the tests for update after the first round of integration test for create is merged in, no worries about that

src/cript/api/api.py
```diff
@@ -514,11 +514,14 @@ def save(self, project: Project) -> None:
...
-            # TODO remove all pre-handled nodes.
+        except CRIPTAPISaveError as exc:
+            if exc.pre_saved_nodes:
+                for node_uuid in exc.pre_saved_nodes:
+                    # TODO remove all pre-saved nodes by their uuid.
+                    pass
             raise exc from exc
 
-    def _internal_save(self, node, known_uuid: Optional[Set[str]] = None) -> Optional[Set[str]]:
+    def _internal_save(self, node, known_uuid: Optional[Set[str]] = None, suppress_attributes: Optional[Dict[str, Set[str]]] = None) -> Optional[Set[str]]:
```
nh916, line 524: can we split up `_internal_save` method into smaller pieces?

I think it is fine for now if we want to keep it to try it out and get it working, but if we can split up into smaller pieces that each do one part of the project then that is all the better because otherwise if we put all the logic for create, update, getting the uuid, etc. into one method it'll make it hard to follow and read.

For me, I try to ideally keep a function/method less than 50 lines of code. Usually when it starts to go over is when I look at the function and can see places that I can split off into their own separate functions. I usually notice that I start losing track of what a function when it's over 50+ lines and lose track of a file when it is 500+ lines.

src/cript/api/utils/save_helper.py
```diff
@@ -1,13 +1,19 @@
...
+def _fix_node_save(api, node, response, known_uuid):
     """
     Helper function, that attempts to fix a bad node.
     And if it is fixable, we resave the entire node.
 
     Returns set of known uuids, if fixable, otherwise False.
     """
-    assert response["code"] in (400, 409)
+    if response["code"] not in (400, 409):
+        raise NotImplementedError(f"The internal helper function `_fix_node_save` has been called for an error that is not yet implemented to be handled {response}.")
```
nh916, line 13: I would just throw a regular exception here because it is likely that this is temporary, that way we don't have to create or think about another exception anywhere reducing our code and making our lives easier

```suggestion
    if response["code"] not in (400, 409):
        raise Exception(f"The internal helper function `_fix_node_save` has been called for an error that is not yet implemented to be handled {response}.")
```

src/cript/api/api.py
InnocentBug: I am doing that already.
All the heavy lifting parts are moved to `save_helpers`.
All the is left in the internal save is the general coordination of which kind of fix is needed when.

src/cript/api/utils/save_helper.py
InnocentBug: `Exception` seems an odd choice though.

`RuntimeError` I can see, and I was debating to use, but `Exception` shouldn't really be used on its own.
(Except for blanket catching of Exceptions.)

src/cript/api/utils/save_helper.py
```diff
@@ -53,3 +62,18 @@ def find_node_by_uuid(node, uuid: str):
...
+def _identify_suppress_attributes(node, response: Dict) -> Optional[Dict[str, Set[str]]]:
+    suppress_attributes = {}
+    if response["error"].startswith("Additional properties are not allowed"):
+        # Find all the attributes, that are listed in the error message with regex
+        attributes = set(re.findall(r"'(.*?)'", response["error"]))
+        path = response["error"][response["error"].rfind("path:") + len("path:") :].strip()
+
+        if path != "/":
+            # TODO find the UUID this belongs to
+            raise NotImplementedError("Fixing non-root objects for patch, not implemented yet")
```
nh916, line 76: I suggest raising a generic exception for this

```suggestion
raise Exception("Fixing non-root objects for patch, not implemented yet")
```

src/cript/api/utils/save_helper.py
nh916: yeah `RuntimeError` probably makes more sense. I usually tend to go with `Exception` whenever I don't have an already created Error and just need a throw away error

InnocentBug:
There is one issue now.

If you actually try to save a node with a duplicate `name` attribute.
The error message you get, might be confusing.

The reason is, that we partially handle this, by removing the `name` from the JSON.
This is necessary to achieve a valid `patch` of nodes with names, but it messes up the regular error too.

Not sure how to fix that right now.

InnocentBug:
This is the script, I have been testing with.
And that works for me, but we need formal testing for this.
(I actually found a little bug because of this.)

```python
import json

import cript

with cript.API(None, None) as api:
    demo_number = "5"

    project = cript.Project(name="My first project. " + demo_number)
    collection = cript.Collection(name="Initial screening")
    # We add this collection to the project as a list.
    project.collection += [collection]
    experiment = cript.Experiment(name="Anionic Polymerization of Styrene with SecBuLi")
    collection.experiment += [experiment]

    toluene = cript.Material(name="toluene" + demo_number, identifiers=[{"smiles": "Cc1ccccc1"}, {"chemical_id": "1140"}])
    styrene = cript.Material(name="styrene" + demo_number, identifiers=[{"smiles": "c1ccccc1C=C"}, {"inchi": "InChI=1S/C8H8/c1-2-8-6-4-3-5-7-8/h2-7H,1H2"}])
    butanol = cript.Material(name="1-butanol" + demo_number, identifiers=[{"smiles": "OCCCC"}, {"inchi_key": "InChIKey=LRHPLDYGYMQRHN-UHFFFAOYSA-N"}])
    methanol = cript.Material(name="methanol" + demo_number, identifiers=[{"smiles": "CO"}, {"names": ["Butan-1-ol", "Butyric alcohol", "Methylolpropane", "n-Butan-1-ol", "methanol"]}])

    inventory = cript.Inventory(
        name="Common chemicals for poly-styrene synthesis",
        material=[toluene, styrene, butanol, methanol],
    )
    collection.inventory += [inventory]

    # # polystyrene = cript.Material(name="my polystyrene", identifiers=[{"bigsmiles": "[H]{[>][<]C(C[>])c1ccccc1[<]}C(C)CC"}])
    polystyrenes = api.search(cript.Material, cript.SearchModes.CONTAINS_NAME, "my polystyrene")
    poly_str = polystyrenes.current_page_results[0]
    polystyrene = cript.load_nodes_from_json(json.dumps(poly_str))
    polystyrene.name += demo_number

    force_field = cript.ComputationalForcefield("opls_aa", "atom")
    polystyrene.computational_forcefield = force_field

    project.material += [polystyrene]

    # api.save(project)

    projects = api.search(cript.Project, cript.SearchModes.CONTAINS_NAME, "My first ")
    for project_str in projects.current_page_results:
        try:
            project = cript.load_nodes_from_json(json.dumps(project_str))
        except cript.nodes.exceptions.CRIPTJsonDeserializationError:
            pass
        if len(project.admin) > 0:
            if project.admin[0].username == "Ludwig Schneider":
                break

    project.name = "asdf"
    api.save(project)
```

nh916:
> This is the script, I have been testing with. And that works for me, but we need formal testing for this. (I actually found a little bug because of this.)
> 
> It might not work for, because you don't have write permissions on the projects.
> 
> ```python
> import json
> 
> import cript
> 
> with cript.API(None, None) as api:
>     demo_number = "5"
> 
>     project = cript.Project(name="My first project. " + demo_number)
>     collection = cript.Collection(name="Initial screening")
>     # We add this collection to the project as a list.
>     project.collection += [collection]
>     experiment = cript.Experiment(name="Anionic Polymerization of Styrene with SecBuLi")
>     collection.experiment += [experiment]
> 
>     toluene = cript.Material(name="toluene" + demo_number, identifiers=[{"smiles": "Cc1ccccc1"}, {"chemical_id": "1140"}])
>     styrene = cript.Material(name="styrene" + demo_number, identifiers=[{"smiles": "c1ccccc1C=C"}, {"inchi": "InChI=1S/C8H8/c1-2-8-6-4-3-5-7-8/h2-7H,1H2"}])
>     butanol = cript.Material(name="1-butanol" + demo_number, identifiers=[{"smiles": "OCCCC"}, {"inchi_key": "InChIKey=LRHPLDYGYMQRHN-UHFFFAOYSA-N"}])
>     methanol = cript.Material(name="methanol" + demo_number, identifiers=[{"smiles": "CO"}, {"names": ["Butan-1-ol", "Butyric alcohol", "Methylolpropane", "n-Butan-1-ol", "methanol"]}])
> 
>     inventory = cript.Inventory(
>         name="Common chemicals for poly-styrene synthesis",
>         material=[toluene, styrene, butanol, methanol],
>     )
>     collection.inventory += [inventory]
> 
>     # # polystyrene = cript.Material(name="my polystyrene", identifiers=[{"bigsmiles": "[H]{[>][<]C(C[>])c1ccccc1[<]}C(C)CC"}])
>     polystyrenes = api.search(cript.Material, cript.SearchModes.CONTAINS_NAME, "my polystyrene")
>     poly_str = polystyrenes.current_page_results[0]
>     polystyrene = cript.load_nodes_from_json(json.dumps(poly_str))
>     polystyrene.name += demo_number
> 
>     force_field = cript.ComputationalForcefield("opls_aa", "atom")
>     polystyrene.computational_forcefield = force_field
> 
>     project.material += [polystyrene]
> 
>     # api.save(project)
> 
>     projects = api.search(cript.Project, cript.SearchModes.CONTAINS_NAME, "My first ")
>     for project_str in projects.current_page_results:
>         try:
>             project = cript.load_nodes_from_json(json.dumps(project_str))
>         except cript.nodes.exceptions.CRIPTJsonDeserializationError:
>             pass
>         if len(project.admin) > 0:
>             if project.admin[0].username == "Ludwig Schneider":
>                 break
> 
>     project.name = "asdf"
>     api.save(project)
> ```

This can be a good demo script or walkthrough for the user with some comments
If you can turn this into an automated test that could be another layer of protection too.

nh916:
@InnocentBug I'm thinking of writing the integration tests today, merging that in, and then merging this into develop so we know the code works for sure. What do you think? or should we merge this into develop and write the tests for it after?

nh916:
> There is one issue now.
> 
> If you actually try to save a node with a duplicate `name` attribute. The error message you get, might be confusing.
> 
> The reason is, that we partially handle this, by removing the `name` from the JSON. This is necessary to achieve a valid `patch` of nodes with names, but it messes up the regular error too.
> 
> Not sure how to fix that right now.

The remedy that I can quickly think of is good documentation to help guide the user through it, but I'll have to work through it a bit to grind off the rough edges and make it smooth.

I wrote a [wiki on errors](https://github.com/C-Accel-CRIPT/Python-SDK/wiki/Error-Messages) that could be helpful to see if it has any good ideas on how to make it better.

InnocentBug:
> This can be a good demo script or walkthrough for the user with some comments
> If you can turn this into an automated test that could be another layer of protection too.

yeah, maybe. I don't know.

InnocentBug:
We'll need to refactor this.

I actually hit the recursion limit on one of my tests.

nh916:
> 

are you thinking of merging in and then refactoring? or refactoring first and then merging in?

src/cript/api/exceptions.py
```diff
@@ -110,10 +110,11 @@ class CRIPTAPISaveError(CRIPTException):
     http_code: str
     api_response: str
 
-    def __init__(self, api_host_domain: str, http_code: str, api_response: str):
+    def __init__(self, api_host_domain: str, http_code: str, api_response: str, pre_saved_nodes: Optional[Set[str]] = None):
```
nh916, line 113: why are we passing in `pre_saved_nodes`?

how come we are passing it in but it's not within the error message?

src/cript/api/utils/save_helper.py
```diff
@@ -1,29 +1,93 @@
...
+        save_values.saved_uuid.add(missing_uuid)
         # Recursive call.
         # Since we should have fixed the "Bad UUID" now, we can try to save the node again
-        return api._internal_save(node, known_uuid)
-    return False
+        return api._internal_save(node, save_values)
+
+    # Handle all duplicate items warnings if possible
+    if response["error"].startswith("duplicate item"):
+        for search_dict_str in re.findall(r"\{(.*?)\}", response["error"]):
```
nh916, line 64: please put an explanation of the regex because they are typically hard to read and understand

src/cript/api/utils/save_helper.py
```diff
@@ -1,29 +1,93 @@
...
+            # The error message contains a description of the offending elements.
+            search_dict_str = "{" + search_dict_str + "}"
+            search_dict_str = search_dict_str.replace("'", '"')
+            search_dict = json.loads(search_dict_str)
+            # These are in the exact format to use with `find_children` so we find all the offending children.
+            all_duplicate_nodes = node.find_children(search_dict)
```
nh916, line 70: we have to upgrade `find_children` to work with arguments instead and possibly enums to make working with it easier because kwargs is harder to write because the user doesn't know all the valid combinations of valid keys and valid values.

src/cript/api/utils/save_helper.py
```diff
@@ -1,29 +1,93 @@
...
+                # After the save, we can reduce it to just a UUID edge in the graph (avoiding the duplicate issues).
+                save_values.saved_uuid.add(str(duplicate_node.uuid))
+
+        return api._internal_save(node, save_values)
+
+    return save_values
 
 
-def get_uuid_from_error_message(error_message: str) -> str:
+def _get_uuid_from_error_message(error_message: str) -> str:
```
nh916, line 90: good change!

src/cript/api/utils/save_helper.py
```diff
@@ -37,10 +101,13 @@ def get_uuid_from_error_message(error_message: str) -> str:
     UUID
         the UUID the API had trouble with
     """
+    bad_uuid = None
     if error_message.startswith("Bad uuid: "):
         bad_uuid = error_message[len("Bad uuid: ") : -len(" provided")].strip()
     if error_message.strip().startswith("Duplicate uuid:"):
         bad_uuid = error_message[len(" Duplicate uuid:") : -len("provided")].strip()
+    if bad_uuid is None or len(bad_uuid) != 36:
```
nh916, line 109: why 36?

src/cript/api/utils/save_helper.py
```diff
@@ -37,10 +101,13 @@ def get_uuid_from_error_message(error_message: str) -> str:
...
+        raise RuntimeError(f"The internal helper function `_get_uuid_from_error_message` has been called for an error message that is not yet implemented to be handled. error message {error_message}, found uuid {bad_uuid}.")
```
nh916, line 110: this reminds me of the [Microsoft task manager story](https://youtu.be/Ve95Nh690l0?t=227) where the developer wrote in his phone number in the error message in case the user encountered

src/cript/api/utils/save_helper.py
```diff
@@ -53,3 +120,18 @@ def find_node_by_uuid(node, uuid: str):
     missing_node = node.find_children({"uuid": uuid})[0]
 
     return missing_node
+
+
+def _identify_suppress_attributes(node, response: Dict) -> Dict[str, Set[str]]:
+    suppress_attributes = {}
+    if response["error"].startswith("Additional properties are not allowed"):
+        # Find all the attributes, that are listed in the error message with regex
+        attributes = set(re.findall(r"'(.*?)'", response["error"]))
```
nh916, line 129: I suggest adding a comment for the regex what it is intending to do

src/cript/api/utils/save_helper.py
```diff
@@ -53,3 +120,18 @@ def find_node_by_uuid(node, uuid: str):
...
+        path = response["error"][response["error"].rfind("path:") + len("path:") :].strip()
```
nh916, line 130: what does this intend to do?

src/cript/api/exceptions.py
InnocentBug: we add this to the exception, because later we can catch this exception and delete all of the pre-saved nodes.
This ensures, that we don't leave orphaned nodes from a failed save attempt.
That is already implemented as far as possible.
(We just need a delete end point.)

nh916: okay I get what you're doing. 
I don't know a way to currently upgrade this, but in the future we will try to see how we can improve it because the code is getting tangled up and will be hard to update one part without another part breaking, but lets just get it working for now and we can come back to it and make it better

src/cript/api/utils/save_helper.py
InnocentBug: It doesn't work with kwargs.

It works with a search dictionary.
A search dictionary, can define nested search criteria.
(For example: `{"node": ["Algorithm"], "parameter": {"name" : "update_frequency"}}`  finds all "Algorithm" nodes, that have a parameter "update_frequency")

You can't do that with another approach, I think.

src/cript/api/utils/save_helper.py
InnocentBug: That is length of a uuid in string.
I can write that better though.

src/cript/api/utils/save_helper.py
InnocentBug: I will explain in the comments.
We are parsing the error message string to get the info we need.

src/cript/api/utils/save_helper.py
nh916: oh I see. you're right that does make it more challenging.

not essential, we'll come back to it later if needed

src/cript/api/utils/save_helper.py
nh916: oh I see,

I think something like this might be helpful, but the current way is also fine with some comments
```python

uuid_str_max_length = 36

if bad_uuid is None or len(bad_uuid) != uuid_str_max_length:
```

src/cript/api/api.py
```diff
@@ -540,33 +540,40 @@ def _internal_save(self, node, save_values: _InternalSaveValues = _InternalSaveV
...
-            # if not successful, we escalate the problem further
-
-        # Handle errors from patching with too many attributes
-        if node_known and response["code"] in (400,):
-            suppress_attributes = _identify_suppress_attributes(node, response)
-            new_save_values = _InternalSaveValues(save_values.saved_uuid, suppress_attributes)
-            save_values += new_save_values
-            return self._internal_save(node, save_values)
+        # Dummy response to have a virtual do-while loop, instead of while loop.
+        response = {"code": -1}
```
nh916, line 544: good choice!

-1 is often associated with an invalid choice, so that is very clear!

nh916:
🏓🏓

src/cript/api/utils/save_helper.py
InnocentBug: I added a comment.

src/cript/api/utils/save_helper.py
InnocentBug: I found a better solution: 
`len(bad_uuid) != len(uuid.uuid4()): # Ensure we found a full UUID describing string (here tested against a random new uuid length.)`

Test the length against the length of an actual UUID.

src/cript/api/utils/save_helper.py
InnocentBug: I added a comment.

src/cript/api/api.py
```diff
@@ -540,33 +540,40 @@ def _internal_save(self, node, save_values: _InternalSaveValues = _InternalSaveV
...
+
+            # Handle errors from patching with too many attributes
+            if node_known and response["code"] in (400,):
+                suppress_attributes = _identify_suppress_attributes(node, response)
+                new_save_values = _InternalSaveValues(save_values.saved_uuid, suppress_attributes)
+                save_values += new_save_values
+
+            # It is only worthwhile repeating the attempted save loop if our state has improved.
+            # Aka we did something to fix the occurring error
+            if not save_values > old_save_values:
```
nh916, line 575: This is clever!

to stop the code from repeating forever

InnocentBug:
It is more iterative now.
There are still recursive calls, but only on different nodes.
That should limit the max recursion depth to max the depth of the graph we are trying to save.

src/cript/api/utils/save_helper.py
```diff
@@ -32,6 +32,25 @@ def __add__(self, other: "_InternalSaveValues") -> "_InternalSaveValues":
                 return_value.suppress_attributes[uuid] = other.suppress_attributes[uuid]
         return return_value
 
+    def __gt__(self, other):
```
nh916, line 36: this makes me a bit worried overriding the `gt` native method, but not sure if I am wrong or not

src/cript/api/utils/save_helper.py
```diff
@@ -75,15 +91,13 @@ def _fix_node_save(api, node, response, save_values: _InternalSaveValues) -> _In
                 try:
                     save_values.suppress_attributes[str(duplicate_node.uuid)] = set(search_dict.keys())
                 except KeyError:
-                    save_values.suppress_attributes[str(duplicate_node.uuid)].add(set(search_dict.keys()))
+                    save_values.suppress_attributes[str(duplicate_node.uuid)].add(set(search_dict.keys()))  # type: ignore
```
nh916, line 95: This needs comments as to what is going on

src/cript/api/utils/save_helper.py
nh916: that is better, that way if the package ever gets updated then our code will still work.

However, it is a bit weird the way that we are generating a new UUID to check that the length are correct. From what I can tell it seems that UUID will always have a length of 36 regardless, so maybe we don't ever have to create a new UUID to check its length against.

src/cript/api/api.py
InnocentBug: Hahaha, yeah I thought so too.

It just needed some limbo make `save_values` comparable.

src/cript/api/utils/save_helper.py
InnocentBug: This is how it's done in python.
Overloading this function allows the use of the `>` operator.

Exactly what we need for the while loop.

And since the code of comparison isn't trivial.
I thought its better to capsule it here.

src/cript/api/utils/save_helper.py
nh916: I would recommend adding a comment for where we got 36 from, otherwise it'll be hard for anyone reading our code to know where we got that from. Either comment or a variable should work.

src/cript/api/utils/save_helper.py
nh916: I think this might need more comments because it could be hard to follow the logic, something like

```
# find all JSON blocks that have `error` in them
for search_dict_str in re.findall(r"\{(.*?)\}", response["error"]):
```

src/cript/api/utils/save_helper.py
InnocentBug: I added some comments.

src/cript/api/utils/save_helper.py
InnocentBug: 36 doesn't appear in the current version of the code anymore.

src/cript/api/utils/save_helper.py
InnocentBug: I have added a comment for this regex.

InnocentBug:
Can we merge this?
Or is there a reason it is idleling. 

I know that 3 tests aren’t passing. But I don’t think they will in the near future. 
we can have a separate Pr for them later

InnocentBug:
The user testing should definitely happen with this branch merged. 

The save function is better and different to the current develop

nh916 (approved):

InnocentBug:
Let's merge when the tests pass

nh916:
> Let's merge when the tests pass

agreed, I think we have 3 tests that have issues, and citation and reference are on the backend and not on us.

we only have parameter to worry about.

the tests that we have are not as rigorous like changing an entire tree, but I think they will do well for a first pass

nh916:
the tests that we have for update are good, but they are very small, in the future we have to be sure to add heftier tests that tests changing a multiple trees within nodes and be sure it passes that test as well

src/cript/api/exceptions.py
```diff
@@ -110,13 +110,21 @@ class CRIPTAPISaveError(CRIPTException):
     http_code: str
     api_response: str
 
-    def __init__(self, api_host_domain: str, http_code: str, api_response: str):
+    def __init__(self, api_host_domain: str, http_code: str, api_response: str, patch_request: bool, pre_saved_nodes: Optional[Set[str]] = None, json_data: Optional[str] = None):
```
nh916, line 113: this exception class will need clean up later

nh916: I recommend passing in the http method as a string or something, we would then manipulate the string to be uppercase so the errors are always uniform

having a bool for it works right now because we only have `POST` and `PATCH`, but in the future when we start to have `DELETE` and `PUT` this will not be flexible

I think it is fine for now to get the project done, but will need refactor in the future to be more maintainable and more friendly

nh916 (approved):

PR #196, opened 2023-07-11
Feature: Download `file.source` that is URL

nh916:
# Description
In the past the SDK could only download files from AWS S3 cloud storage, but now it has the rough ability to download files from URL as well.

## Known Issues
* if the URL needs JS enabled, needs sessions, has extra needs, then the downloader throws an error
* the error needs to be tested to be sure that it is user friendly
* refactoring needed: https://trello.com/c/qLaN2hqR

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -639,6 +639,7 @@ class API:
         # return the object_name within AWS S3 for easy retrieval
         return object_name
 
+    @beartype
     def download_file(self, object_name: str, destination_path: str = ".") -> None:
         """
         download a file from AWS S3 and save it to the specified path on local storage
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -22,6 +22,7 @@ from cript.api.exceptions import (
 from cript.api.paginator import Paginator
 from cript.api.utils.get_host_token import resolve_host_and_token
 from cript.api.utils.save_helper import fix_node_save
+from cript.api.utils.web_file_downloader import download_file_from_url
 from cript.api.valid_search_modes import SearchModes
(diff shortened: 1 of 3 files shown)


src/cript/api/api.py
```diff
@@ -675,6 +677,11 @@ def download_file(self, object_name: str, destination_path: str = ".") -> None:
             just downloads the file to the specified path
         """
 
+        # if the file source is a URL
+        if object_name.startswith("http"):
```
InnocentBug, line 681: I am a little confused, don't AWS S3 files also start with `http`?

nh916: nope no longer.

the way I used to retrieve the objects from the AWS S3 cloud storage was to convert what I just uploaded into a link and save that into the file source. However, I later found that the file link expires after a day or so. 

The way that both frontend and SDK grabs the files from AWS S3 is by getting them via their object_names such as `/python_sdk_files/my_uploaded_file.txt`

so now the file node source can be:
**file node local file source:** `/python_sdk_files/my_uploaded_file.txt`
**file node url source:** `https://my-file-link.pdf`

InnocentBug: see `file.py` `is_local`

nh916: oooh you are right!

should we merge this in because it is about downloading web file source and then tackle the refactoring in a separate PR?

InnocentBug: Up to you, I am not a big fan of merging broken code, but sometimes that is necessary.

InnocentBug (dismissed):

InnocentBug (approved):

PR #197, opened 2023-07-11
Integration Tests for Update/PATCH

nh916:
# Description
Integration tests for update functionality
within each node's test file, the `test_integration...` handles that node's create and update

## Changes
* after create is done within the integration tests, then I change a small attribute on the desired node, to trigger an update
    * In most places I've tried putting `"UPDATE"` within the changed attribute, but for some I could not do that
* changed print statements formatting to make it easier to see and not get lost within the terminal outputs
* All the create tests passed fine and I added a a little separator, and then the update tests run after create tests separator is:
```
######################################## TEST Passed ########################################
```

## Notes
* The code for comparing the JSONs is not perfect, because I have noticed a couple of errors that have gotten through, however, I think is good enough for now and we can upgrade it in the future. 
    * it only checks the fields that the JSONs have in common from what I can tell and the regex that I have written ignores `UID` since that will always be different between the JSONs
    * I have spent a lot of time on this before, and every solution was either awful code or just did not compare the correct things. I think this is the best I can figure out for now. 

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -581,7 +581,7 @@ class API:
 
         if response["code"] != 200:
             print(node_known, node.get_json(known_uuid=save_values.saved_uuid, suppress_attributes=save_values.suppress_attributes, indent=2).json)
-            raise CRIPTAPISaveError(api_host_domain=self._host, http_code=response["code"], api_response=response["error"], pre_saved_nodes=save_values.saved_uuid)  # type: ignore
+            raise CRIPTAPISaveError(api_host_domain=self._host, http_code=response["code"], api_response=response["error"], pre_saved_nodes=save_values.saved_uuid, json_data=json_data)  # type: ignore
 
         save_values.saved_uuid.add(str(node.uuid))
         return save_values
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -540,33 +540,40 @@ class API:
         for file_node in node.find_children({"node": ["File"]}):
             file_node.ensure_uploaded(api=self)
 
-        # We assemble the JSON to be saved to back end.
-        # Note how we exclude pre-saved uuid nodes.
-        json_data = node.get_json(known_uuid=save_values.saved_uuid, suppress_attributes=save_values.suppress_attributes).json
+        # Dummy response to have a virtual do-while loop, instead of while loop.
+        response = {"code": -1}
 
-        # This checks if the current node exists on the back end.
-        # if it does exist we use `patch` if it doesn't `post`.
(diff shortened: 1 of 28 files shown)


nh916:
@InnocentBug I wrote the update integration tests so that there is a simple attribute changed to trigger update. I thought this would be easiest when debugging and working with it.

I think a simpler way that we could also go is just changing the notes for every primary node to `"my notes UPDATED"` that way we can always quickly check if notes has been updated or not without having to look through the JSON. I thought that might be too simple, so I didn't do that, but another route that we can go. 

I could also start adding several changes for update if we're feeling confident to see how several attributes of a node can be changed and if it'll be okay or not

nh916:
I am happy to put this either in the [handle-patch branch](https://github.com/C-Accel-CRIPT/Python-SDK/pull/193) or [develop branch](https://github.com/C-Accel-CRIPT/Python-SDK/tree/develop)

InnocentBug:
> I am happy to put this either in the [handle-patch branch](https://github.com/C-Accel-CRIPT/Python-SDK/pull/193) or [develop branch](https://github.com/C-Accel-CRIPT/Python-SDK/tree/develop)

I think it should target the `handle-patch` branch.

(But we can still merge that PR first, if we feel like.

nh916:
> > I am happy to put this either in the [handle-patch branch](https://github.com/C-Accel-CRIPT/Python-SDK/pull/193) or [develop branch](https://github.com/C-Accel-CRIPT/Python-SDK/tree/develop)
> 
> I think it should target the `handle-patch` branch.
> 
> (But we can still merge that PR first, if we feel like.

Is that PR done? I thought we wanted it to handle updates, but the update tests are currently failing

I would suggest
1. merge this into handle_patch
3. after handle_patch can update and all update integration tests pass
4. then we merge it into develop

what are your thoughts?

InnocentBug:
> > > I am happy to put this either in the [handle-patch branch](https://github.com/C-Accel-CRIPT/Python-SDK/pull/193) or [develop branch](https://github.com/C-Accel-CRIPT/Python-SDK/tree/develop)
> > 
> > 
> > I think it should target the `handle-patch` branch.
> > (But we can still merge that PR first, if we feel like.
> 
> Is that PR done? I thought we wanted it to handle updates, but the update tests are currently failing
> 
> I would suggest
> 
> 1. merge this into handle_patch
> 2. after handle_patch can update and all update integration tests pass
> 3. then we merge it into develop
> 
> what are your thoughts?

We might want to consider this done. 
idk. It depends on how fixable we think the problems are. 

We can merge and figure out the rest later.

InnocentBug (approved):

PR #202, opened 2023-07-14
API Token and Storage Token Update

nh916:
# Description
frontend now allows users to grab http token and cloud storage from security page. This update is changed to reflect working with that.

## Changes
* renamed `token` to `api_token` to differentiate between different kinds of tokens
* putting `Bearer` in front of the token within the code and user can just pass in a raw JWT and we prepare it for them
* updated reading from config.json and environment variables accordingly
    * updated example script docs as well to use the new format
* added `storage_token` to interact with AWS S3 in cognito mode
* changed `docs/exampls/synthesis.md` api initialization to pass the CI
    * if we want to have them blank and read env vars automatically, then we'll have to alter our tests.yaml workflow that I would rather not do in this PR

## Tests
* the file upload and download test worked well
* interacting with the API worked well
* tests that need to be skipped are no longer commented out, but are using [pytest skip](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skipping-test-functions)
* wrote test for initializing api with `None`
* > not every different way to initialize the SDK has a test written so there is a chance that if the user mixes the configurations there could be issues
* > more testing is needed because there are several options and something could clash or something could break with one of the options

## Known Issues
* lack of testing covering all scenarios for the initialization
* having multiple options for initialization can lead to bugs that are harder to track down and fix


InnocentBug (changes requested):
The tests aren’t passing yet. 

It is a working design, I think but not great. 

How does the front end get the storage token?
Is there an end point where it is requested from?

ideally we would not have the user provide the storage token separately. 

Instead we request that token like the front end. And validate that request with the api token as before.

InnocentBug (approved):

PR #208, opened 2023-07-19
making file extension mandatory on file.py

nh916:
# Description

## Changes
* made file node mandatory and every file node must have an extension
* updating tests that were not updated after refactoring of saving file during save instead of during script
* making file extension mandatory on file.py
* during download gets file name from file node instead of requiring it from the user

## Tests
* updated the tests to work with the new style of uploading during save
* using pytest.skip to ignore tests that cannot be ran during CI
* 

## Checklist


InnocentBug:
Close for now, we have other prios now.

PR #211, opened 2023-07-20
updated db schema to validate with `POST` and `PATCH`

nh916:
# Description
Previously the db schema only ran against `NodePost`, but with this method it will work with `NodePatch` as well

## Changes
* added an optional Boolean flag in `cript.API._is_node_schema_valid` of `is_patch` and if it is then it will use the `NodePatch` and if not then it will use a `NodePost`

## Tests
* updated API DB Schema tests to work correctly with the new optional parameter

## Known Issues
* if there will also be a `NodeDelete` in the future, then we will need to update it
    * in that instance, I will opt for an enum and allow users to choose from that
* need tests for PATCH schema in the future
    * ticket: https://trello.com/c/P1KCbHZ6

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -437,7 +437,7 @@ class API:
             return self._db_schema
 
     @beartype
-    def _is_node_schema_valid(self, node_json: str) -> bool:
+    def _is_node_schema_valid(self, node_json: str, is_patch: bool = False) -> bool:
         """
         checks a node JSON schema against the db schema to return if it is valid or not.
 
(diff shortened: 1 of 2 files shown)


nh916:
@InnocentBug we discussed `adding a flag to get_json for if it is post or patch, if patch validate against patch instead of post`

however, validation happens within the `json property` so a flag in `get_json` wouldn't do anything for validation and the property does not take arguments.

Putting it inside of `get_json` will create a stack overflow error. 

how do you want to validate the JSON against POST or PATCH because json property would have to know that it is a post or patch at some point?

InnocentBug (approved):

InnocentBug:
> I can see adding a flag for `get_json` if it is a PATCH so that it is serialized differently

yes, that is what we need.
validate is not going to be call from inside `get_json` that doesn't work as it creates an infinite call loop.

nh916:
@InnocentBug I see this has already been merged into `handle-patch` should I just close this PR?

InnocentBug:
sure that would work.
Let's merge that branch though too.

nh916:
this is already merged into [handle-patch](https://github.com/C-Accel-CRIPT/Python-SDK/tree/handle-patch)

PR #219, opened 2023-07-21
Upgrade: Importing Fixtures Automatically into conftest.py

nh916:
# Description
import all fixtures in primary, subobject, and supporting nodes. This way it is imported automatically instead of being imported one at a time and if we make a new fixture, we don't have to remember to import it into conftest.py, as it will automatically get imported in

## Changes
* importing fixtures in with wildcard instead of one at a time

## Tests
* tested a few nodes to be sure there is no fixture or path error and they worked fine

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,6 @@ The fixtures are all functional fixtures that stay consistent between all tests.
 """
 
 import pytest
-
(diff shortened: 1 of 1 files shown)


InnocentBug (approved):
not a superfan of wild card imports, but it works I guess.

PR #221, opened 2023-07-24
UX: Node Validation Output

nh916:
# Description
Putting in a log statement in the package to show that the program is working in the background as the user is waiting for it to load.

## Changes
* wrote an INFO log statement to show things are happening in the background and it is not just standing still
* wrote an option to turn off the log statements if you don't want to see them
  * using the variable name of `verbose` because it is standard and well known that it is responsible for terminal logs

### Terminal Output
image

### Documentation
image

## Checklist

### DIFF
--- a/.trunk/configs/.cspell.json
+++ b/.trunk/configs/.cspell.json
@@ -99,6 +99,7 @@
     "openmm",
     "equi",
     "Navid",
-    "ipykernel"
+    "ipykernel",
+    "levelname"
   ]
 }
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -1,5 +1,6 @@
 import copy
 import json
+import logging
 import os
 import uuid
 import warnings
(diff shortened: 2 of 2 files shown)


InnocentBug:
can we capsule this a little bit. Have a log level, that only shows it when the log level is high enough.
And also give some more context, at least print out the type of node, so the user sees that something changes while validating.

There definitely needs to be an option to turn it off.

nh916:
> can we capsule this a little bit. Have a log level, that only shows it when the log level is high enough. And also give some more context, at least print out the type of node, so the user sees that something changes while validating.
> 
> There definitely needs to be an option to turn it off.

done! ✅ 

I was trying to use logs yesterday and it kept not showing up in the terminal for hours, until I eventually gave up and used print statements. Tried it again today and it worked fine, so I guess something was stuck in ram yesterday.

PR #222, opened 2023-07-24
added `is_local_file` boolean flag to File node

nh916:
# Description
added `is_local_file` boolean flag to File node so the user can specify if their file is local or not manually

## Changes
* this effects the constructor to file node during creation
 * the source property setter is replaced with a function

## Tests
 * changed tests for it to work well, however, there is an issue where deep_copy will not work for the file node
 * did not run an entire complete test, but just ran the file tests to be sure it is working well

## Checklist

### DIFF
--- a/docs/examples/simulation.md
+++ b/docs/examples/simulation.md
@@ -244,10 +244,10 @@ experiment.computation += [init, equilibration, bulk, ana]
 New we'd like to upload files associated with our simulation. First, we'll instantiate our File nodes under a specific project.
 
 ```python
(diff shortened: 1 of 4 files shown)


nh916:
@InnocentBug I don't really like this approach. Nothing against it, but it just doesn't feel very appealing.

nh916:
> for the UX, we probably want the default for `is_local` to be True, (also in tutorials).
> 
> I think we can add some more finesse to this approach.

I think either way True or False works for me. What do you mean by finesse? what are you thinking?

You're thinking of putting this in the next release right?

InnocentBug:
file node needs some refactoring anyways.

PR #224, opened 2023-07-25
wrote test function for `_is_node_field_valid()`

nh916:
# Description
this function was missing a test case and since I am thinking of refactoring it soon, test cases would be really helpful with that and being sure that the project works well anyways

## Changes
* wrote test function and test cases for `_is_node_field_valid()`
* made a change that if the field is a list, but the string inside is empty then it is invalid

## Checklist

### DIFF
--- a/src/cript/nodes/util/__init__.py
+++ b/src/cript/nodes/util/__init__.py
@@ -286,7 +286,7 @@ def _is_node_field_valid(node_type_list: list) -> bool:
     """
 
     # TODO consider having exception handling for the dict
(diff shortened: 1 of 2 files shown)


nh916:
@InnocentBug I was thinking of wrapping the function with a `@beartype` so that it could give us errors in case we give it a wrong type of input at some point within our code.  What are your thoughts?

InnocentBug (approved):

InnocentBug:
this is excellent

nh916:
> this is excellent

Thank you so much!

If you can think of more test cases feel free to add them.

PR #236, opened 2023-07-28
fix the deep diff comparison.

InnocentBug:
This now fails a bunch of tests.

(Feel free to add the `skip` to pytest.
I don't know how to do that.

More importantly, when locally run it shows all the issues of the back end.

### DIFF
--- a/src/cript/nodes/util/__init__.py
+++ b/src/cript/nodes/util/__init__.py
@@ -286,7 +286,7 @@ def _is_node_field_valid(node_type_list: list) -> bool:
     """
 
     # TODO consider having exception handling for the dict
-    if isinstance(node_type_list, list) and len(node_type_list) == 1 and isinstance(node_type_list[0], str) and node_type_list[0]:
+    if isinstance(node_type_list, list) and len(node_type_list) == 1 and isinstance(node_type_list[0], str) and len(node_type_list[0]) > 0:
         return True
     else:
         return False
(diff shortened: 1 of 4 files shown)


nh916:
I'll run this locally and get a print out of the integration tests

src/cript/nodes/util/__init__.py
```diff
@@ -286,7 +286,7 @@ def _is_node_field_valid(node_type_list: list) -> bool:
     """
 
     # TODO consider having exception handling for the dict
-    if isinstance(node_type_list, list) and len(node_type_list) == 1 and isinstance(node_type_list[0], str) and node_type_list[0]:
+    if isinstance(node_type_list, list) and len(node_type_list) == 1 and isinstance(node_type_list[0], str) and len(node_type_list[0]) > 0:
```
nh916, line 289: I would write this instead as `node_type_list[0]` because if it is full then it would be truthy and if empty it is false

tests/api/test_api.py
```diff
@@ -237,7 +237,7 @@ def test_download_file_from_url(cript_api: cript.API, tmp_path) -> None:
     # the path it will save it to will be `tmp_path/downloaded_file_name.json`
     path_to_save_file: Path = tmp_path / "downloaded_file_name"
 
-    cript_api.download_file(file_source=url_to_download_file, destination_path=str(path_to_save_file))
+    cript_api.download_file(url_to_download_file, str(path_to_save_file))
```
nh916, line 240: I think this should work

```py
cript_api.download_file(file_source=url_to_download_file, destination_path=str(path_to_save_file))
```

I don't understand why it would break

nh916 (approved):

src/cript/nodes/util/__init__.py
InnocentBug: It’s testing if the string isn’t enpty

PR #237, opened 2023-07-28
Tests integration with `ON/OFF` switch

nh916:
# Description
Test integration works much easier now and can be switched ON/OFF with a boolean and any test that is skipped is marked transparently in the output

so now if we want to run integration tests we can easily switch `HAS_INTEGRATION_TESTS_ENABLED=False` to True and we are done

## Changes
* added ON/OFF boolean variable in conftest.py to easily switch them ON/OFF
* added if statement to skip integration tests to make skipping them easy and transparent
* added skip decorator to unused file tests to make skipping them transparent
* uncommented a file test that works well currently
* cleaned up `test_user.py`

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,7 +17,7 @@ from fixtures.supporting_nodes import *
 import cript
 
 
-HAS_INTEGRATION_TESTS_ENABLED: bool = False
+HAS_INTEGRATION_TESTS_ENABLED: bool = True
 
 
 @pytest.fixture(scope="session", autouse=True)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,7 +16,6 @@ from fixtures.supporting_nodes import *
 
 import cript
 
-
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -16,6 +16,8 @@
 
 import cript
 
+HAS_INTEGRATION_TESTS_ENABLED: bool = False
```
InnocentBug, line 19: This is great. 
much needed automation of this. 

One small thing. Maybe the default here is True. But we load and env variable that can override it to false. And in the automated tests we set that variable. 

The advantage being, that no code has to be changed to run the full tests locally.

InnocentBug (approved):
Coverage fails because the percentage isn’t reached. Which makes sense. Should be ok to lower it for this.

tests/conftest.py
nh916: yeah I can put that in a new PR and we can always set it to `True` or `False` ourselves to override it and make testing easy if needed

PR #238, opened 2023-07-29
added vocabulary links to all nodes in documentation

nh916:
# Description
added vocabulary links to documentation of where users can get the vocab for each attribute.

## Notes
If I could make the site url a variable and refer to it instead of having to hardcode it, then that would be the best because I could just change the domain to production and I am simply done. But I could not figure that out.

`mkdocs.yaml`
```yaml
site_domain: https://mycriptapp.org
```

`docstring`
```
[my link]({{ site_domain}}/about/)
```

I tried Jinja, [macro plugin](https://mkdocs-macros-plugin.readthedocs.io/en/latest/) but I was not able to get it to work

I ended up [leaving a question on mkdocs discussion board](https://github.com/squidfunk/mkdocs-material/discussions/5775) for now and will hopefully get an answer soon.

Looked through stackoverflow and found no answers

## Checklist

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -49,6 +49,8 @@ nav:
       - API Exceptions: exceptions/api_exceptions.md
       - Node Exceptions: exceptions/node_exceptions.md
   - FAQ: faq.md
+  - Internal Wiki Documentation: https://github.com/C-Accel-CRIPT/Python-SDK/wiki
+  - CRIPT Python SDK Discussions: https://github.com/C-Accel-CRIPT/Python-SDK/discussions
 
 theme:
   name: material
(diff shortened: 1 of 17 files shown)


nh916:
@InnocentBug let me know if you are able to figure out how to make the domain into a variable that we can dynamically load into the documentation dynamically instead of hard coding. I tried it for a bit and could not get it to work and then just did it statically.

https://mkdocs-macros-plugin.readthedocs.io/en/latest/

https://github.com/squidfunk/mkdocs-material/discussions/5775

InnocentBug (approved):
I see the issue here, linking to mycriptapp.org is problematic.

But I don't know how to make it better :(

nh916:
> I see the issue here, linking to mycriptapp.org is problematic.
> 
> But I don't know how to make it better :(

I think if I keep messing around with mkdocs and [macros](https://mkdocs-macros-plugin.readthedocs.io/en/latest/), I might be able to set a variable for the domain and pull that into the documentation and append to the string. 

Therefore, when we want to change to production, I only have to change one variable.

PR #240, opened 2023-07-31
renamed `integration_test_helper` file

nh916:
# Description
renamed `integration_test_helper` file because it is not a test, but a helper/utility file

## Changes
* renamed all occurrences of it throughout code

## Checklist


nh916:
not sure if this is a good idea or not because it is technically doing a test, but it is not a test itself as it cannot be ran and is just a utility function. I think renaming it to `helper file` might prevent any wrong assumptions that it is a test

InnocentBug (approved):
this is a great change.
This is indeed not running a test itself, just providing helper functions for other tests, so this is a better name.

InnocentBug (approved):
this is a great change.
This is indeed not running a test itself, just providing helper functions for other tests, so this is a better name.

PR #242, opened 2023-07-31
change test coverage requirement from 90% to 89%

nh916:
# Description
reducing coverage requirement to 89% to not get failures in CI.

## Notes
We are actually over `90%` test coverage, but since the CI skips integration tests, it comes out to less than `90%`

## Checklist


nh916:
I think we have tests covering almost everything, so it would be hard to create another test to bump up the test coverage.

I thought of creating a test for pagiantor.py class, but since that will send a request to the API to get results, it will also need to be skipped and will not bump up the overall test coverage percentage.

InnocentBug (approved):

PR #243, opened 2023-08-01
refactor enum name from `ControlledVocabularyCategories` to `VocabCategories`

nh916:
# Description
I think the enum class name of `ControlledVocabularyCategories` is just too long and makes the code look less clean I think. I think if we change it to a shorter like `VocabCategories` it will still create the same meaning for the user and keep the code clean and concise.

## Changes
* changed documentation in controlled_vocabulary_categories.md
* changed example documentation
* changed how it is imported into __init__.py
* changed the test_api.py for it
* changed how it is used in api.py
* changed how it is used in material_deserialization.py

## Notes
I think for the most accurate results, we would have to ask users about it what they thought and if they liked it we keep it, otherwise we remove it. I think this is something that is large and I am guessing users will either want it shorter or will not miss it if it becomes shorter.

## Screenshots
image

## Checklist

### DIFF
--- a/docs/api/controlled_vocabulary_categories.md
+++ b/docs/api/controlled_vocabulary_categories.md
@@ -1,1 +1,1 @@
-::: cript.ControlledVocabularyCategories
+::: cript.VocabCategories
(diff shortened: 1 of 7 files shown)


nh916:
I am also happy to keep this until after the first release, so we are not doing any big refactoring before the release

InnocentBug (approved):

PR #245, opened 2023-08-01
updating requirements and putting `jupytext` into `requirements_docs.txt`

nh916:
# Description
updating requirements and putting `jupytext` into `requirements_docs.txt` because it is only used within documentation

## Checklist

### DIFF
--- a/.github/workflows/test_examples.yml
+++ b/.github/workflows/test_examples.yml
@@ -19,20 +19,26 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: [3.11]
+
     env:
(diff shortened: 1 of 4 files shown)


InnocentBug (approved):

InnocentBug (changes requested):
doesn't work, because loads `requirements_dev.txt` not `requirements_docs.txt`

InnocentBug (approved):
this should work now.

nh916:
@InnocentBug we could also add this workflow within the `build and deploy docs` workflow as well, since I think it would only require a couple of steps for it to be in there and functioning

PR #247, opened 2023-08-02
Wrote documentation for `load_nodes_from_json`

nh916:
# Description
Wrote documentation for `load_nodes_from_json` as that will be useful for the user as they are working through the code.

## Screenshots
image

## Checklist

### DIFF
--- a/src/cript/nodes/util/__init__.py
+++ b/src/cript/nodes/util/__init__.py
@@ -294,7 +294,7 @@ def _is_node_field_valid(node_type_list: list) -> bool:
 
 def load_nodes_from_json(nodes_json: str):
     """
-    User facing function, that return a node and all its children from a json input.
+    User facing function, that return a node and all its children from a json string input.
 
     Parameters
     ----------
--- a/src/cript/nodes/util/__init__.py
+++ b/src/cript/nodes/util/__init__.py
@@ -339,7 +339,7 @@ def load_nodes_from_json(nodes_json: str):
 
     Returns
     -------
-    CRIPT Node or List of CRIPT Nodes
+    Union[CRIPT Node, List[CRIPT Node]]
         Typically returns a single CRIPT node,
         but if given a list of nodes, then it will serialize them and return a list of CRIPT nodes
     """
--- a/src/cript/nodes/util/__init__.py
+++ b/src/cript/nodes/util/__init__.py
@@ -19,7 +19,6 @@ from cript.nodes.exceptions import (
 )
 from cript.nodes.primary_nodes.experiment import Experiment
 from cript.nodes.primary_nodes.project import Project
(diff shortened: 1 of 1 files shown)


nh916:
@InnocentBug I wrote the documentation from the best I could figure out, let me know if it is wrong or missing, you know this code better than I do.

src/cript/nodes/util/__init__.py
```diff
@@ -292,9 +293,54 @@ def _is_node_field_valid(node_type_list: list) -> bool:
         return False
 
 
-def load_nodes_from_json(nodes_json: str):
+def load_nodes_from_json(nodes_json: str) -> UUIDBaseNode:
```
InnocentBug, line 296: This type hint is only kind of true.

Suppose your JSON is a list of nodes, like
```
[
 { node: ["Material"], ...},
 {node: ["Collection"], ...},
 ...
 ]
 ```
 
 Then you would get a list with the types `List[cript.Material, cript.Collection, ...]`.
 And this will work for they two container types JSON supports, list and dict. 

So the function is a bit more flexible then the type hint suggests.

Comment (InnocentBug) [reply src/cript/nodes/util/__init__.py:296]: I don't think we use that functionality anywhere yet though.

File: src/cript/nodes/util/__init__.py [hunk 2]
```diff
@@ -292,9 +293,54 @@ def _is_node_field_valid(node_type_list: list) -> bool:
...
+    This function uses a custom `_NodeDecoderHook` to convert JSON nodes into Python objects.
+    The `_NodeDecoderHook` class is responsible for handling the deserialization of nodes
+    and caching objects with shared UIDs to avoid redundant deserialization.
+
+    The function is intended for deserializing CRIPT nodes and should not be used for generic JSON.
+
+
+    Returns
+    -------
+    UUIDBaseNode
```
InnocentBug, line 342: see comment above

src/cript/nodes/util/__init__.py
nh916: hmm I see what you're saying, but not sure how to resolve it other than putting all nodes in a union for the return type. I can also happily just leave it out and try to explain in the docstrings. What are your thoughts?

InnocentBug: yeah, as long as we don't enforce it with @beartype I am ok with that.

InnocentBug: or, just remove it and explain the the docs.

nh916: I made changes to the documentation, let me know if it is satisfactory or not. Feel free to also update the docs as well if you can quickly word it better.

InnocentBug (approved):

PR #248, opened 2023-08-02
added documentation for `_NodeDecoderHook`

nh916:
# Description
wrote some documentation for `_NodeDecoderHook`

## Checklist

### DIFF
--- a/src/cript/nodes/util/__init__.py
+++ b/src/cript/nodes/util/__init__.py
@@ -155,6 +155,23 @@ class _UIDProxy:
 
 class _NodeDecoderHook:
     def __init__(self, uid_cache: Optional[Dict] = None):
+        """
+        Initialize the custom JSON object hook used for CRIPT node deserialization.
+
+        Parameters
+        ----------
+        uid_cache : Optional[Dict], optional
+            A dictionary to cache Python objects with shared UIDs, by default None.
+
+        Notes
+        -----
+        The `_NodeDecoderHook` class is used as an object hook for JSON deserialization,
+        handling the conversion of JSON nodes into Python objects based on their node types.
+        The `uid_cache` is an optional dictionary to store cached objects with shared UIDs
+        to avoid unnecessary deserialization and circular references.
+
+
+        """
         if uid_cache is None:
             uid_cache = {}
         self._uid_cache = uid_cache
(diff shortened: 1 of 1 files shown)


nh916:
@InnocentBug I wrote the documentation from the best I could figure out, let me know if it is wrong or missing, you know this code better than I do.

src/cript/nodes/util/__init__.py
```diff
@@ -155,6 +155,23 @@ def __init__(self, uid: str):
...
+        Parameters
+        ----------
+        uid_cache : Optional[Dict], optional
+            A dictionary to cache Python objects with shared UIDs, by default None.
+
+        Notes
+        -----
+        The `_NodeDecoderHook` class is used as an object hook for JSON deserialization,
+        handling the conversion of JSON nodes into Python objects based on their node types.
+        The `uid_cache` is an optional dictionary to store cached objects with shared UIDs
```
InnocentBug, line 170: The uid cache is not to avoid loops.

It is to never create two different python objects with the same uid.

So if you pass a JSON for a node with a UID that is present in the uid_cache, instead of creating a new python object for that node, the existing python object from the `uid_cache` is used and updated with the values from the JSON.

src/cript/nodes/util/__init__.py
```diff
@@ -175,7 +192,26 @@ def __call__(self, node_str: Union[dict, str]) -> dict:
...
+
+        Parameters
+        ----------
+        node_str : Union[dict, str]
+            The JSON representation of a node or a regular dictionary.
+
+        Returns
+        -------
+        dict
+            A Python dictionary representing the deserialized node or the input dictionary itself.
```
InnocentBug, line 205: It returns a regular dictionary if the input JSON or input dict is not a node.

If it is a node, it returns a the appropriate node object.
For example, if you give it a JSON of Material node, it will return an object of type `cript.Material`.

InnocentBug (changes requested):

src/cript/nodes/util/__init__.py
nh916: updated it

src/cript/nodes/util/__init__.py
nh916: updated it

InnocentBug (approved):

PR #249, opened 2023-08-02
Wrote documentation for `NodeEncoder` class

nh916:
# Description
Wrote documentation for `NodeEncoder` class

## Screenshots
image

## Checklist

### DIFF
--- a/src/cript/nodes/util/__init__.py
+++ b/src/cript/nodes/util/__init__.py
@@ -194,7 +194,7 @@ class NodeEncoder(json.JSONEncoder):
         nodes_to_condense = serialize_dict["node"]
         for node_type in nodes_to_condense:
             if node_type in self.condense_to_uuid:
-                attributes_to_process = self.condense_to_uuid[node_type]
+                attributes_to_process = self.condense_to_uuid[node_type]    # type: ignore
                 for attribute in attributes_to_process:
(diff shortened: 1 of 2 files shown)


src/cript/nodes/util/__init__.py
```diff
@@ -22,12 +22,80 @@
...
+    condensed representations to avoid redundancy in the JSON output.
+    It also allows suppressing specific attributes from being included in the serialized JSON.
+
+    Attributes
+    ----------
+    handled_ids : Set[str]
+        A set to store the UIDs of nodes that have been processed during serialization.
+    known_uuid : Set[str]
+        A set to store the UUIDs of nodes that have been previously encountered in the JSON.
+    condense_to_uuid : Set[str]
```
InnocentBug, line 38: Ok, I think this is my fault.

This isn't actually the type of `condense_to_uuid` the actual type is `Dict[str, List[str]]`, with the key being the name of the node, and the value a list (maybe we should make it a Set) of the attributes.

nh916: whatever you think is best works with me

InnocentBug: i fixed it

nh916: Thank you so much!

PR #253, opened 2023-08-02
conftest env var upgrade

nh916:
# Description
small suggestion for conftest.py

## Checklist


InnocentBug:
This doesn't work for me.

If no variable is specified it skips tests.
The opposite should be the case.

nh916:
> This doesn't work for me.
> 
> If no variable is specified it skips tests. The opposite should be the case.

yeah I thought that is what you were going for, but give me one sec and I can flip it to true if no variable is specified

InnocentBug (approved):

nh916:
this implementation really isn't my favorite. the previous one was much more cleaner and if the developer doesn't have the env variable then they get an error and they fix it and they move on, but we can keep it for now

PR #255, opened 2023-08-04
HTTP `DELETE` node

nh916:
# Description
delete a node from the API and locally

## Changes
* moved logger to the top of the file to declare the format for the whole file 
* wrote delete method
* wrote delete docstrings

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -49,8 +49,10 @@ def _get_global_cached_api():
         raise CRIPTAPIRequiredError()
     return _global_cached_api
 
+
 # Set the logging configuration for the file
 logging.basicConfig(format="%(levelname)s: %(message)s", level=logging.INFO)
+
 
 class API:
     """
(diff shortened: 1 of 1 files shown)


InnocentBug (approved):

nh916:
deleting it from API is pretty easy I would assume, but doing it locally can be more challenging
here's what I'm thinking, we have to:
1. get the node (e.g. material node) and project that they want to delete it from
2. traverse the project node to get to that material node
3. within the material node, we need to have a pointer to the parent node
    1. we must also know what the parent node is from all of its other attributes
4. go up to the parent node, and set that attribute to `None` that the user specified
5. however, if the node is required to be there, then we have to throw an error and let the user deal with it

---

I think an easier way to do this, could be that 

1. we just sent an HTTP DELETE request to the API
2. After getting HTTP 200
3. We get the project node from API
4. Deserialize it and overwrite the local project node and not have to manually handle delete
    1. this requires that the API give back the correct data in its response

I think I prefer the second way

PR #256, opened 2023-08-04
Conftest setup for multiple environments

nh916:
# Description
Setup conftest.py with multiple config.json files to easily jump between server environments

this would require the existence of several config files within the `tests/` directory
* development_config.json
* staging_config.json
* production_config.json

each of which data inside of it that looks like this
```json
{
  "host": "https://development.api.mycriptapp.org/",
  "api_token": "development api token",
  "storage_token": "development storage token"
}
```

## Changes
* making api client with config.json
    * using file path and path object for it to work correctly regardless of from which directory the code was ran
* updated CI to work correctly
    * removed unneeded env vars such as `CRIPT_HOST`, `CRIPT_TOKEN`, `CRIPT_STORAGE_TOKEN`
* added config files to .gitignore so they are not accidentally uploaded to the repo

## Notes
* for more clarity we can also give a log in the terminal of which environment is being used in case the developer is bumping into bugs and forgot to change their environment. That would give them a small reminder of what environment they are in.
* I wrote this in multiple ways and all felt like over engineering, so I wrote it this way, which is the cleanest and least amount of code I can write
    * > for more please look in the git history
* I think this would make working with multiple environments quick and easy
    * Any developer jumping on the project can also quickly setup their config file and jump into the project quickly

## Checklist


nh916:
@InnocentBug let me know if this works with you and your thoughts. I can work on the CI if this is something that we want

InnocentBug:
I don't know. Maybe this is a good idea.
As soon as we have proper CI/CD this needs to go though.

These tokens should be handled as github secrets.
So, not sure what to do about it.

I don't really like commited config files for this, but I can see where you are coming from.

nh916:
> I don't know. Maybe this is a good idea. As soon as we have proper CI/CD this needs to go though.
> 
> These tokens should be handled as github secrets. So, not sure what to do about it.
> 
> I don't really like commited config files for this, but I can see where you are coming from.

I think those are fair points.
I can close this PR for now and we can come back to it in the future if jumping between cript instances becomes too cumbersome

PR #261, opened 2023-08-08
Made optional attributes optional in setters type hinting

nh916:
# Description

## Checklist

### DIFF
--- a/src/cript/nodes/primary_nodes/collection.py
+++ b/src/cript/nodes/primary_nodes/collection.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field, replace
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Union
 
 from beartype import beartype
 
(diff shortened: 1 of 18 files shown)


nh916:
@InnocentBug what are your thoughts on this?

nh916:
lists cannot be optional and instead the user must give an empty list to wipe that field. This must be communicated clearly in the documentation so the user is not left wondering

InnocentBug:
> @InnocentBug what are your thoughts on this?

I think that is OK.
Not highest priority if you ask me though.

InnocentBug:
> lists cannot be optional and instead the user must give an empty list to wipe that field. This must be communicated clearly in the documentation so the user is not left wondering

not entirely true.
we can handle this like in the constructors.
If a None gets passed in the the function (the optional part), we convert it after into an empty list.

nh916:
implemented everything for this PR better in this [PR](https://github.com/C-Accel-CRIPT/Python-SDK/pull/280) and it is already merged, so closing this one

PR #262, opened 2023-08-09
Std args json

InnocentBug:
# Description

This rehandles the std fields in JSON.

1) it add default fields when they are not sent by the API. This helps to tolerate missing required arguments.
2) default fields are better suppressed when serializing, this helps to reduce false positives during integration testing.

## Notes
This required to change some default list fields from `None` to list.
## Checklist

### DIFF
--- a/src/cript/nodes/primary_nodes/collection.py
+++ b/src/cript/nodes/primary_nodes/collection.py
@@ -59,10 +59,10 @@ class Collection(PrimaryBaseNode):
         # TODO add proper typing in future, using Any for now to avoid circular import error
         member: List[User] = field(default_factory=list)
         admin: List[User] = field(default_factory=list)
-        experiment: Optional[List[Any]] = None
-        inventory: Optional[List[Any]] = None
+        experiment: List[Any] = field(default_factory=list)
+        inventory: List[Any] = field(default_factory=list)
         doi: str = ""
-        citation: Optional[List[Any]] = None
+        citation: List[Any] = field(default_factory=list)
 
     _json_attrs: JsonAttributes = JsonAttributes()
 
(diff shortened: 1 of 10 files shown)


src/cript/nodes/primary_nodes/collection.py
```diff
@@ -59,10 +59,10 @@ class JsonAttributes(PrimaryBaseNode.JsonAttributes):
         # TODO add proper typing in future, using Any for now to avoid circular import error
         member: List[User] = field(default_factory=list)
         admin: List[User] = field(default_factory=list)
-        experiment: Optional[List[Any]] = None
-        inventory: Optional[List[Any]] = None
+        experiment: List[Any] = field(default_factory=list)
```
nh916, line 62: oh I see, so by default you gave it a list in case not provided by API

nh916:
3 thoughts

1. is this safe to add before ACS?
2. are there any known errors that could come up?
3. where are you getting the default fields for the node?

src/cript/nodes/primary_nodes/collection.py
InnocentBug: Yes, exactly

InnocentBug:
> 3 thoughts
> 
> 1. is this safe to add before ACS?
> 2. are there any known errors that could come up?
> 3. where are you getting the default fields for the node?

1) yes. Once we fixed the bugs in the tests. It actually will help, because it makes pulling nodes from api more easy. Less error for users
2). It that I know of. The failing tests is just that we am have to adjust the test data we compare to
3) we get the default fields from the default values of the data classes. At some point it would be ideal to set those with the help of the db schema. But for now they are hard coded.

nh916:
tested this locally against staging, and it was good to go! ✅

nh916 (approved):

nh916 (approved):

PR #268, opened 2023-08-12
skipping `cript.API.search UUID` for now

nh916:
# Description
need to set up the UUID test for all 3 servers later after ACS launch, no time to change it currently 

## Checklist


nh916:
fixed this for the new tests where I get a node via exact name, get the UUID from the node I just got, then search the API via UUID. So it is kind of a test of both `exact name` and `search UUID`

I think it works correctly for now

PR #273, opened 2023-08-15
Temp no orphan

InnocentBug:
# Description

Do not merge!

This is just a temporary thing.

## Changes

This allows you to disable the checks.

If you just want to disable the project orphan checks do this:

```python
with cript.API(None, None, None) as api:
   api.no_orphan_checks = True
   #...
```

You can also disable DB schema tests in general:

```python
with cript.API(None, None, None) as api:
   api.no_schema_checks = True
   #...
```


InnocentBug:
formally implemented.

PR #280, opened 2023-08-18
test/develop optional node attributes can be removed

nh916:
# Description
removing optional attributes from nodes to be sure that it works correctly

## Changes
* refactored and clean up some tests
* removed `copy.deep_copy` to make tests simpler
* renamed some variable names from `c2` to `complex_condition_node` to be easier to read and more self documenting
    * added logical spacing between some tests to be easier to read

## Known Issues
* `test_condition.py` was having issues with removal of `Data` node, which [data model](https://pubs.acs.org/doi/suppl/10.1021/acscentsci.3c00011/suppl_file/oc3c00011_si_001.pdf) says that it should be optional, but when removing it the DB Schema gives me an error
* `test_software_configuration` is missing some tests in `test_setter_getter`
   * the tests will be added in future PR

## Checklist

### DIFF
--- a/src/cript/nodes/primary_nodes/material.py
+++ b/src/cript/nodes/primary_nodes/material.py
@@ -272,7 +272,7 @@ class Material(PrimaryBaseNode):
 
     @parent_material.setter
     @beartype
-    def parent_material(self, new_parent_material: "Material") -> None:
+    def parent_material(self, new_parent_material: Optional["Material"]) -> None:
         """
         set the [parent materials](./) for this material
 
--- a/src/cript/nodes/subobjects/condition.py
+++ b/src/cript/nodes/subobjects/condition.py
@@ -365,7 +365,7 @@ class Condition(UUIDBaseNode):
         return self._json_attrs.uncertainty
 
     @beartype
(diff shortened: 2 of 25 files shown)


InnocentBug (approved):

src/cript/nodes/primary_nodes/material.py
```diff
@@ -272,7 +272,7 @@ def parent_material(self) -> Optional["Material"]:
 
     @parent_material.setter
     @beartype
-    def parent_material(self, new_parent_material: "Material") -> None:
+    def parent_material(self, new_parent_material: Optional["Material"]) -> None:
```
nh916, line 275: @InnocentBug would it be okay if this is allowed to be removed later in the script?

InnocentBug: From the scientific standpoint probably no.
Once a material is derived from another material, that won't change.

However, from a practical standpoint, it does make sense.
A parent material might have been added by mistake, or we copy this material (to make a new one) and that doesn't have parent.

So, I would leave it in.

nh916: okie dokes, I agree with you, that is a very good point

PR #282, opened 2023-08-21
Pointing file upload tests to staging AWS S3 storage

nh916:
# Description
changed cognito variables for API in tests to point to staging storage

## Changes
* changed conftest to override production API object class variables
* spaced out code a bit to be easier to read

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -49,8 +49,17 @@ def cript_api():
     storage_token = os.getenv("CRIPT_STORAGE_TOKEN")
 
     assert cript.api.api._global_cached_api is None
+
     with cript.API(host=None, api_token=None, storage_token=storage_token) as api:
+
+        # overriding AWS S3 cognito variables to be sure we do not upload test data to production storage
+        # staging AWS S3 cognito storage variables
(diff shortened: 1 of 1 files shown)


InnocentBug (approved):

tests/conftest.py
```diff
@@ -49,8 +49,16 @@ def cript_api():
     storage_token = os.getenv("CRIPT_STORAGE_TOKEN")
 
     assert cript.api.api._global_cached_api is None
+
     with cript.API(host=None, api_token=None, storage_token=storage_token) as api:
+        # overriding AWS S3 cognito variables to be sure we do not upload test data to production storage
+        # staging AWS S3 cognito storage variables
+        api._IDENTITY_POOL_ID = "us-east-1:25043452-a922-43af-b8a6-7e938a9e55c1"
```
InnocentBug, line 56: Ideally, we make an end-point for the API at one point, that delivers these variables.

nh916: I have not worked with AWS S3 cognito before so I am unsure of the convention, workflow, security, or if it has access to these variables. However, even if they are hardcoded in a config file in the API and sent to the SDK, that could be more dynamic

PR #283, opened 2023-08-21
raising error for getting db schema via requests

nh916:
# Description
gives slightly better error if the user is out of USA

#### Before
```bash
       try:
            return complexjson.loads(self.text, **kwargs)
        except JSONDecodeError as e:
            # Catch JSON-related errors and raise as requests.JSONDecodeError
            # This aliases json.JSONDecodeError and simplejson.JSONDecodeError
>           raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
E           requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

..\..\venv\lib\site-packages\requests\models.py:975: JSONDecodeError
```
> this error is because when the API is pinged from outside of USA it returns a text of `"You must be in the US to access Cript app."` and since this is not JSON it gives a bad JSON deserialization error

#### After
After raising error from requests, the error becomes
```bash
        if http_error_msg:
>           raise HTTPError(http_error_msg, response=self)
E           requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://api.criptapp.org/api/v1/schema/

..\..\venv\lib\site-packages\requests\models.py:1021: HTTPError
```

The error is slightly better but I would have preferred `"wrong URL for db schema"` and `"Cannot use CRIPT outside of USA"`

I think this new error should be good enough for now and if we get users complaining then we can change it

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -517,11 +517,14 @@ class API:
 
         # fetch db_schema from API
         else:
-            # fetch db schema, get the JSON body of it, and get the data of that JSON
-            response = requests.get(url=f"{self.host}/schema/").json()
+            # fetch db schema from API
+            response = requests.get(url=f"{self.host}/schema/")
 
-            if response["code"] != 200:
-                raise APIError(api_error=response.json())
+            # raise error if not HTTP 200
+            response.raise_for_status()
+
+            # if no error, take the JSON from the API response
+            response = response.json()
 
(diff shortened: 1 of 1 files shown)


nh916:
waiting on this [response](https://mit-olsenlab.slack.com/archives/C04DVDLHWHH/p1692662530546349)

InnocentBug (approved):

PR #287, opened 2023-08-22
Refactor/Renamed material `identifiers` to `identifier`

nh916:
# Description
One of the users made a comment that every other attribute on a node is singular except for identifiers which is plural and it would be more intuitive if it was singular

## Changes
* changed all files, tests, and documentation to use `identifier` instead of `identifiers`
  * all tests are passing
* fixed any broken links for vocabulary within documentation
* removed unneeded test for test_file.py

## Tests
* all tests are passing after heavy refactoring

## Checklist

### DIFF
--- a/src/cript/nodes/supporting_nodes/file.py
+++ b/src/cript/nodes/supporting_nodes/file.py
@@ -94,7 +94,7 @@ class File(PrimaryBaseNode):
     | Attribute       | Type | Example                                                                                               | Description                                                                 | Required |
     |-----------------|------|-------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------|----------|
     | source          | str  | `"path/to/my/file"` or `"https://en.wikipedia.org/wiki/Simplified_molecular-input_line-entry_system"` | source to the file can be URL or local path                                 | True     |
-    | type            | str  | `"logs"`                                                                                              | Pick from [CRIPT File Types](https://app.criptapp.org/vocab/file-type/)          | True     |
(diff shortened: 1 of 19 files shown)


src/cript/nodes/supporting_nodes/file.py
```diff
@@ -94,7 +94,7 @@ class File(PrimaryBaseNode):
     | Attribute       | Type | Example                                                                                               | Description                                                                 | Required |
     |-----------------|------|-------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------|----------|
     | source          | str  | `"path/to/my/file"` or `"https://en.wikipedia.org/wiki/Simplified_molecular-input_line-entry_system"` | source to the file can be URL or local path                                 | True     |
-    | type            | str  | `"logs"`                                                                                              | Pick from [CRIPT File Types](https://app.criptapp.org/vocab/file-type/)          | True     |
+    | type            | str  | `"logs"`                                                                                              | Pick from [CRIPT File Types](https://app.criptapp.org/vocab/file_type/)          | True     |
```
InnocentBug, line 97: there are couple of these changes that chane `-` to `_`.
Is that intentional?

InnocentBug (approved):

src/cript/nodes/supporting_nodes/file.py
nh916: yep, I was using [Xenu](https://home.snafu.de/tilman/xenulink.html#Download) to catch any broken links, but it doesn't do well when it tries to go on CRIPT UI, so those were less caught. Recently, I found that the old frontend had urls as `file-type` and the new one has them as `file_type` and I fixed any that I could find

PR #293, opened 2023-08-25
Feature:  HTTP `DELETE` node

nh916:
# Description
giving users a way to delete a nodes from the API

### Wrote Delete
* wrote `cript.API.delete()` method
* wrote integration tests for delete
* wrote full documentation for delete

---

### Updated Logger
* created a `set_log()` method
  * it essentially sets the logger for the whole `cript.API` class
    * was considering making it into a getter that would create a new logger and return it, but since each class should only have one logger that might encourage bad programming and create bugs
  * made `verbose` into a property that changes the class logger after the user sets it.
    * The user simply sets it casually and we rearrange the logger behind the scenes for them to work correctly so they can turn verbose `ON` and `OFF` easily throughout their script 
    * The logger has good simple logic so we don't have to keep putting `if self.verbose then log`
  * I am not sure if the way I am doing logger is perfect right now, but I think it is good for our use cases for now and we can upgrade it as we go
* wrote/updated documentation for `cript.API.verbose` and the logging in general for that class
---

### Updated Paginator Errors
* changed paginator exception from `Exception` to `APIError`
* upgraded/refactored `APIError`
  * `APIError` did not have as much information and was unhelpful in debugging
  * gave it more information and made it much more helpful for users to read and for us to debug

---

### Updated Documentation
* updated documentation: noticed that documentation for `cript.API` tells users to use host as `criptapp.org` and since that was outdated, I changed it to `api.criptapp.org`

## Tests
* wrote out `integrate_delete_node_helper`

### Terminal Log: `cript.API.delete()` Successfully Deleted a Node
This is logged to the terminal after the SDK successfully deletes a node from the API
image

### New APIError Exception Terminal Output:
image

### Documentation: `cript.API.delete()`
screencapture-127-0-0-1-8000-api-api-2023-08-29-18_40_401

### Documentation: `cript.API logging`
image

## Known Issues
* We are only using logger in the `cript.API` but if we start putting it in more places then we'll need a refactor for it to make it cleaner and consider putting it within `cript/logger.py`

### Integration Tests
**Recommendation:** I recommend running the all SDK tests locally to confirm everything I did works correctly

<details>
    <summary> click to see screenshots of all tests ran locally </summary>
        
</details>

## Checklist

### DIFF
--- a/src/cript/api/exceptions.py
+++ b/src/cript/api/exceptions.py
@@ -1,4 +1,4 @@
-from typing import List, Set, Optional
+from typing import List, Optional, Set
 
 from cript.exceptions import CRIPTException
 
--- a/src/cript/api/exceptions.py
+++ b/src/cript/api/exceptions.py
@@ -184,11 +184,28 @@ class APIError(CRIPTException):
 
     api_error: str = ""
 
-    def __init__(self, api_error: str) -> None:
+    # having the URL that the API gave an error for helps in debugging
+    api_url: Optional[str] = None
+    http_method: Optional[str] = None
+
+    def __init__(self, api_error: str, http_method: Optional[str] = None, api_url: Optional[str] = None) -> None:
         self.api_error = api_error
 
+        self.api_url = api_url
+
+        # TODO consider having an enum for all the HTTP methods so they are easily entered and disallows anything
+        #   that would be not make sense
+        self.http_method = http_method
+
     def __str__(self) -> str:
-        error_message: str = f"The API responded with {self.api_error}"
+        # TODO refactor all of SDK using this error to be used the same to avoid this logic and optional attributes
+        # this logic currently exists to make previous SDK work
+
+        # if you have the URL then display it, otherwise just show the error message
+        if self.api_url and self.http_method:
+            error_message: str = (f"CRIPT Python SDK sent HTTP `{self.http_method.upper()}` request to URL: `{self.api_url}` and API responded with {self.api_error}")
+        else:
+            error_message: str = f"The API responded with: {self.api_error}"
 
         return error_message
 
--- a/src/cript/api/exceptions.py
+++ b/src/cript/api/exceptions.py
@@ -203,11 +203,9 @@ class APIError(CRIPTException):
 
         # if you have the URL then display it, otherwise just show the error message
         if self.api_url and self.http_method:
-            error_message: str = (f"CRIPT Python SDK sent HTTP `{self.http_method.upper()}` request to URL: `{self.api_url}` and API responded with {self.api_error}")
+            return f"CRIPT Python SDK sent HTTP `{self.http_method.upper()}` request to URL: `{self.api_url}` and API responded with {self.api_error}"
         else:
-            error_message: str = f"The API responded with: {self.api_error}"
-
-        return error_message
+            return f"The API responded with: {self.api_error}"
 
 
 class FileDownloadError(CRIPTException):
--- a/src/cript/api/exceptions.py
+++ b/src/cript/api/exceptions.py
@@ -1,5 +1,5 @@
 import json
-from typing import List, Set, Optional
+from typing import List, Optional, Set
 
 from cript.exceptions import CRIPTException
 
--- a/src/cript/api/exceptions.py
+++ b/src/cript/api/exceptions.py
@@ -1,4 +1,4 @@
-from typing import List, Optional, Set
+from typing import List, Set, Optional
 
 from cript.exceptions import CRIPTException
 
@@ -198,7 +198,7 @@ class APIError(CRIPTException):
         self.http_method = http_method
 
     def __str__(self) -> str:
-        # TODO refactor all of SDK using this error to be used the same to avoid this logic and optional attributes
+        # TODO refactor all of SDK using this error to be used the same to avoid this logic and Optional attributes
         # this logic currently exists to make previous SDK work
 
         # if you have the URL then display it, otherwise just show the error message
--- a/src/cript/api/exceptions.py
+++ b/src/cript/api/exceptions.py
@@ -1,5 +1,5 @@
 import json
-from typing import List, Set
+from typing import List, Set, Optional
 
 from cript.exceptions import CRIPTException
 
--- a/tests/integration_test_helper.py
+++ b/tests/integration_test_helper.py
@@ -144,10 +144,13 @@ def delete_integration_node_helper(cript_api: cript.API, node_to_delete: UUIDBas
     cript_api.delete(node=node_to_delete)
 
     # should not be able to get node by UUID anymore because it is deleted and API should return an  error
-    with pytest.raises(APIError):
-        cript_api.search(
-            node_type=node_to_delete,
-            search_mode=cript.SearchModes.UUID,
-            value_to_search=str(node_to_delete.uuid)
-        )
 
+    # with pytest.raises(APIError):
+    deleted_node_paginator = cript_api.search(
+        node_type=node_to_delete,
+        search_mode=cript.SearchModes.UUID,
+        value_to_search=str(node_to_delete.uuid)
+    )
+
+    # be sure API responded with empty results
+    assert len(deleted_node_paginator.current_page_results) == 0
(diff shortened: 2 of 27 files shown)


nh916:
I wrote delete before here: https://github.com/C-Accel-CRIPT/Python-SDK/pull/255

I'll be putting that logic into this branch and much more. Then I'll delete the old branch.

nh916:
@InnocentBug for `Project`, `Material`, and `Property` the API gives bad error messages instead of giving `404`

I can think of a few ways to go about it:
1. wait for API to fix their errors
  * leave the tests as they are and wait for the API to fix the errors so getting deleted nodes gives good consistent errors
  * once the API errors are fixed the tests will run correctly as expected
2. fix the tests now based on what we have
  * fix the delete tests so that it expects the current error from the API
3. assume API will never change the errors and refactor the SDK delete to account for the current errors

what are your thoughts on this?

#### More context

For all nodes, when trying to get a node by UUID the API responds with:
```json
{
   "code":404,
   "data":"None",
   "error":"The requested URL was not found on the
   server. If you entered the URL manually please
   check your spelling and try again."
}
```

For `Project` `Material` and `Property` it responds with
```json
{
   "code":401,
   "data":"None",
   "error":"The server could not verify that you are authorized to access the URL requested. You either supplied the wrong credentials (e.g. a bad password), or your browser doesn't understand how to supply the credentials required."
}
```

InnocentBug:
> fix the tests now based on what we have
> 
>     * fix the delete tests so that it expects the current error from the API

Hmm, that is a tough one.

In the past we always chose 2.

But I am not really the right person to make these calls, you can ping PM with this if you want.
You really nicely laid out what the decision options are.

nh916:
> > fix the tests now based on what we have
> > ```
> > * fix the delete tests so that it expects the current error from the API
> > ```
> 
> Hmm, that is a tough one.
> 
> In the past we always chose 2.
> 
> But I am not really the right person to make these calls, you can ping PM with this if you want. You really nicely laid out what the decision options are.

Thank you so much! 😅

I can put this on our slack channel

nh916:
@InnocentBug any ideas on what is making CI unhappy?

it seems to think that `Optional` is not imported into `src/cript/api/exceptions.py` but it is imported and when I run mypy locally it runs successfully.

InnocentBug:
> @InnocentBug any ideas on what is making CI unhappy?
> 
> it seems to think that `Optional` is not imported into `src/cript/api/exceptions.py` but it is imported and when I run mypy locally it runs successfully.

Yeah, weird one.
No idea where it comes from.

I got it happy by upgrading trunk #301

nh916:
> I was originally in a bit of a dilemma when first writing this but I think to keep the whole documentation consistent I will keep Numpy style for the warnings. Since I already wrote out the whole post, I'll post it anyways so we are all aware.

## SDK Warning

The warning could look a bit better if we decide to go with either [material mkdocs admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/?h=admon#supported-types) or change the whole project to use to [Google style docstrings](https://google.github.io/styleguide/pyguide.html) as [MKDocstrings](https://mkdocstrings.github.io/) seems to parse [Google style docstrings](https://google.github.io/styleguide/pyguide.html) style better.

> I have left a [discussion post on MKDocstrings repo better parsing Numpy style](https://github.com/mkdocstrings/mkdocstrings/discussions/564#discussioncomment-6512690), but I don't think much came of it 

### Numpy Style
[Numpy documentation](https://numpydoc.readthedocs.io/en/latest/format.html#warns) way of writing a warning is like this

#### Docstrings Code
```python
"""
Warnings
--------
After successfully deleting a node from the API, keep in mind that your local Project node in your script
may still contain outdated data as it has not been synced with the API.

To ensure you have the latest data, follow these steps:

1. Fetch the newest Project node from the API using the [`cript.API.search()`](./#cript.api.api.API.search) provided by the SDK.
1. Deserialize the retrieved data into a new Project node using the [`load_nodes_from_json`](../../utility_functions/#cript.nodes.util.load_nodes_from_json) utility function.
1. Replace your old Project node with the new one in your script for accurate and up-to-date information.
"""
```

#### Documentation Output
image

#### IDE Documentation Output
image


### Material MKDocs Style
However, we can use [material mkdocs admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/?h=admon#supported-types) to style it up a bit and make it look better

#### Docstrings Code
```python
"""
!!! Warning "Refresh Project Node After Deletion"
	After successfully deleting a node from the API, keep in mind that your local Project node in your script
	may still contain outdated data as it has not been synced with the API.

	To ensure you have the latest data, follow these steps:

	1. Fetch the newest Project node from the API using the [`cript.API.search()`](./#cript.api.api.API.search) provided by the SDK.
	1. Deserialize the retrieved data into a new Project node using the [`load_nodes_from_json`](../../utility_functions/#cript.nodes.util.load_nodes_from_json) utility function.
	1. Replace your old Project node with the new one in your script for accurate and up-to-date information.
"""
```

#### Documentation Output
image

#### IDE Documentation Output
image

nh916:
This is a bit of a big PR. I apologize for that.

This could be broken down into smaller PRs if we need, but in the interest of making our lives easy and not having merge conflicts I think we should keep it as one PR

src/cript/api/exceptions.py
```diff
@@ -214,13 +214,28 @@ class APIError(CRIPTException):
...
+        #   that would be not make sense
+        self.http_method = http_method
+
+    def __str__(self) -> str:
+        # TODO refactor all of SDK using this error to be used the same to avoid this logic and Optional attributes
+        # this logic currently exists to make previous SDK work
+
+        # if you have the URL then display it, otherwise just show the error message
+        if self.api_url and self.http_method:
+            return f"CRIPT Python SDK sent HTTP `{self.http_method.upper()}` request to URL: `{self.api_url}` and API responded with {self.api_error}"
```
InnocentBug, line 236: You allow self.http_method to be None.
So you can't call `upper` on it.

tests/integration_test_helper.py
```diff
@@ -96,3 +97,47 @@ def integrate_nodes_helper(cript_api: cript.API, project_node: cript.Project):
...
+    > within search we can also add assert statements to be sure that the API gave the same node
+    > that we searched for. Not needed at the time of writing this test
+    """
+
+    if not HAS_INTEGRATION_TESTS_ENABLED:
+        pytest.skip("DELETE integration tests with API requires real API and Storage token")
+        return
+
+    # check that the node we want to delete first exists on the API
+    my_node_paginator = cript_api.search(
```
InnocentBug, line 126: I understand the logic.
But would it be better to just send `delete` and handle the error from API if the node doesn't exist?

InnocentBug (changes requested):
I have some minor requests.
Plus I think coverage is failing now.

tests/integration_test_helper.py
nh916: that could streamline the process, but the only thing that we would lose from that would be that we did not prove that the node was there before we went to delete it, so if we start getting bugs we might be asking ourselves, 

`"is it having issues because the API can't delete the node? or is it because the node never existed on the API?"`

I was thinking that writing it this way could be a bit more explicit, but removing the first search would honestly not take away much as far as I can see and would be pretty easy too.

You think remove it?

src/cript/api/exceptions.py
nh916: That is a great point! I'll change it to be an empty string

nh916: Well, I am first checking to see if it exists, and only if it does, then I call upper on it. If it is `None` then it would hit the else

InnocentBug: Oh, I missing that. My bad.

tests/integration_test_helper.py
InnocentBug: I think, I would insist on getting rid of it in production code, but this is tests only and should be OK.

InnocentBug (approved):

tests/integration_test_helper.py
nh916: I would honestly be happy getting rid of it too either now or in a future PR. I think both should work and be okay

PR #294, opened 2023-08-28
Support a limited number of nodes per POST/PATCH request

InnocentBug:
# Description

The idea here is to count the number of nodes during the production of the JSON.
Unfortunately, we have to hack the `json` `NodeEncoder` class even more to do this.

We save now a set of all nodes being fully expressed in the JSON.

If that number exceeds the accepted number nodes per request, we save randomly one node from the graph first.
After this presaving, we show this node only as UUID in the graph and try again. 

## Changes
It gets more complicated and hard to read.

## Tests

We don't have a test for this yet.
https://trello.com/c/wWoz8l5l

## Known Issues

It doesn't work yet currently because of this:
https://trello.com/c/kjNVBX40

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -618,7 +618,8 @@ class API:
             Just sends a `POST` or `Patch` request to the API
         """
         try:
-            self._internal_save(project, max_number_of_nodes_per_save=4)
+            # TODO: replace the hard coded 100 with a request to the API end point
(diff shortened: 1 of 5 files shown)


InnocentBug:
Yeah, so this doesn't work.
Waiting for clarification of the back end devs, on how to do this.

InnocentBug:
I don't think this will work.
Probably postpone until, @brili has an idea of save can be fixed.

PR #295, opened 2023-08-28
Fix duplicated name error msg

InnocentBug:
# Description
We knew, and Berenger pointed it out in his testing: our error message for duplicated names was garbage.

So, this is a first attempt to fix this.
## Changes

If a node name is now duplicated you get this error now:
```
    raise CRIPTDuplicateNameError(response, json_data, exc) from exc
cript.api.exceptions.CRIPTDuplicateNameError: The name My first project. 7 for your Project node is already present in CRIPT.
 Either choose a new name, or consider namespaces like 'MyNameSpace::My first project. 7 instead.
```
Instead of 
```
cript.api.exceptions.CRIPTAPISaveError: API responded to POST with 'http:400 'name' is a required property at path: /' data: {"node": ["Material"], "uid": "_:c85c6d44-7c5d-4f32-87d9-fa8834079345", "uuid": "c85c6d44-7c5d-4f32-87d9-fa8834079345", "computational_forcefield": {"node": ["ComputationalForcefield"], "uid": "_:c9f8491b-0b0a-4f50-8c1b-0e5f490c97fc", "uuid": "c9f8491b-0b0a-4f50-8c1b-0e5f490c97fc", "key": "opls_aa", "building_block": "atom"}, "bigsmiles": "[H]{[>][<]C(C[>])c1ccccc1[<]}C(C)CC"}
```

## Tests
None added yet, we may want to consider one.

## Known Issues
We rely on the exact *undocumented* repsonse from the backend.
If that fails, you would only get the old message though.

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -683,7 +683,17 @@ class API:
             # If we get an error we may be able to fix, we to handle this extra and save the bad node first.
             # Errors with this code, may be fixable
             if response["code"] in (400, 409):
-                returned_save_values = _fix_node_save(self, node, response, save_values)
+                try:
+                    returned_save_values = _fix_node_save(self, node, response, save_values)
+                except CRIPTAPISaveError as exc:
+                    # If the previous error was a duplicated name issue (first if condition)
+                    # And (second condition) the request failed bc of the now suppressed name
+                    if "duplicate item [{'name':" in response["error"] and "'name' is a required property" in exc.api_response:
+                        # Raise a save error, with the nice name related error message
+                        raise CRIPTAPISaveError(api_host_domain=exc.api_host_domain, http_code=response["code"], api_response=response["error"], patch_request=exc.patch_request, pre_saved_nodes=exc.pre_saved_nodes, json_data=json_data) from exc
+                    # Else just raise the exception as normal.
+                    raise exc
+
                 save_values += returned_save_values
 
             # Handle errors from patching with too many attributes
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -16,6 +16,7 @@ from cript.api.exceptions import (
     CRIPTAPIRequiredError,
     CRIPTAPISaveError,
     CRIPTConnectionError,
+    CRIPTDuplicateNameError,
     InvalidHostError,
     InvalidVocabulary,
 )
@@ -690,7 +691,7 @@ class API:
                     # And (second condition) the request failed bc of the now suppressed name
                     if "duplicate item [{'name':" in response["error"] and "'name' is a required property" in exc.api_response:
                         # Raise a save error, with the nice name related error message
-                        raise CRIPTAPISaveError(api_host_domain=exc.api_host_domain, http_code=response["code"], api_response=response["error"], patch_request=exc.patch_request, pre_saved_nodes=exc.pre_saved_nodes, json_data=json_data) from exc
+                        raise CRIPTDuplicateNameError(response, json_data, exc) from exc
                     # Else just raise the exception as normal.
                     raise exc
 
--- a/src/cript/api/exceptions.py
+++ b/src/cript/api/exceptions.py
@@ -1,4 +1,5 @@
-from typing import List, Optional, Set
+import json
+from typing import List, Set
 
 from cript.exceptions import CRIPTException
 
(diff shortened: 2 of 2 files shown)


InnocentBug:
@bearmit for visibility

nh916:
```
consider namespaces like 'MyNameSpace::My first project. 7' instead.
```

this is a bad idea to tell users to do this if this is not the officially agreed upon idea for the platform by the whole team. If this is agreed upon by the whole team and has been communicated well to team then we can do do it. We just don't want to mislead or have users create bad data that later becomes a headache for the team to work with and fix.

src/cript/api/api.py
```diff
@@ -683,7 +684,18 @@ def _internal_save(self, node, save_values: Optional[_InternalSaveValues] = None
...
             # Errors with this code, may be fixable
             if response["code"] in (400, 409):
-                returned_save_values = _fix_node_save(self, node, response, save_values)
+                try:
+                    returned_save_values = _fix_node_save(self, node, response, save_values)
+                except CRIPTAPISaveError as exc:
+                    # If the previous error was a duplicated name issue
+                    if "duplicate item [{'name':" in str(response["error"]):
+                        # And (second condition) the request failed bc of the now suppressed name
+                        if "'name' is a required property" in exc.api_response:
```
nh916, line 693: I wonder if this would be cleaner if we remove the nested if and combine them into an if statement with `and`

src/cript/api/exceptions.py
```diff
@@ -110,7 +111,7 @@ class CRIPTAPISaveError(CRIPTException):
     http_code: str
     api_response: str
 
-    def __init__(self, api_host_domain: str, http_code: str, api_response: str, patch_request: bool, pre_saved_nodes: Optional[Set[str]] = None, json_data: Optional[str] = None):
+    def __init__(self, api_host_domain: str, http_code: str, api_response: str, patch_request: bool, pre_saved_nodes: Set[str], json_data: str):
```
nh916, line 114: I am okay approving this, but just concerned that if our error messages get too big it will be really hard to read and debug for users and ourselves.

We can also tackle cleaning up our errors after we have all features done

src/cript/api/exceptions.py
```diff
@@ -129,6 +130,35 @@ def __str__(self) -> str:
...
+            self.name = "unknown_name_key"
+        try:
+            self.node = json_dict["node"][0]
+        except KeyError:
+            self.node = "UnknownTypeKey"
+        except IndexError:
+            self.node = "UnknownTypeIdx"
+
+    def __str__(self) -> str:
+        return f"The name '{self.name}' for your {self.node} node is already present in CRIPT. Either choose a new name!"  # , or consider namespaces like 'MyNameSpace::{self.name}' instead."
```
nh916, line 159: Good message!

Tells the user exactly what node had issues and how they can easily solve it!

src/cript/api/exceptions.py
```diff
@@ -129,6 +130,35 @@ def __str__(self) -> str:
...
+        try:
+            json_dict = json.loads(self.json_data)
+        except (TypeError, json.JSONDecodeError):
+            self.name = "unknown_name"
+            self.node = "UnknownType"
+        try:
+            self.name = json_dict["name"]
+        except KeyError:
+            self.name = "unknown_name_key"
+        try:
```
nh916, line 151: I wonder if there is a cleaner way of writing this or not

nh916 (approved):

nh916:
I do agree that this is a better error message that is more helpful to us and users

src/cript/api/exceptions.py
InnocentBug: The error message doesn't actually get longer.
The contrary, the message itself is shorter now.

There is a fundamental reason for requiring all this information.
This extra info can be used to clean up save calls that go wrong.
For example we have the infrastructure to delete trees that got only saved halfway through.
We were just waiting for `delete` to implement it.

And by inheriting and passing this info down, this capability remains intact even for NameErrors during saving, but with an improved error message.

src/cript/api/exceptions.py
InnocentBug: Yes, there is.

You could use blanket style exception catching like this:

```python
try:
  json_dict = json.loads(self.json_data)
  self.name = json_dict["name"]
  self.node = json_dict["node"][0]
except (TypeError, json.JSONDecodeError, KeyError, IndexError):
  self.name = "unknown_name"
  self.node = "UnknownType"
```

But that is less granular and might not be the best possible error message.

Let me know what you prefer.

src/cript/api/exceptions.py
InnocentBug: thx :)

src/cript/api/api.py
nh916: fair point, agreed!

src/cript/api/exceptions.py
nh916: hmm that is true. We can leave it how it was then for now and if we find a better way in the future we can refactor and upgrade it then. It has good comments so no one should be lost as to what is happening I think

PR #296, opened 2023-08-28
100+ nodes project for testing

nh916:
# Description
created a `BIG Project node` that has 100+ nodes in it for testing

## Notes
The code is written in a NOT DRY way, so later  we will need to do slight refactoring to clean it up, and possibly a helper function that creates all the needed nodes for the `test_api_node_limit` test function. We can do this in our clean up period. I think it will work for now.

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -662,6 +662,7 @@ class API:
 
             # If the number of nodes is larger then accepted by the backend.
             # Try to save individual nodes first.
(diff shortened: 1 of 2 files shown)


InnocentBug:
thx for the test @nh916.
It does exactly what we needed.

It fails with my implementation though.
No idea why, I guess the backend counts nodes differently then I do.

PR #299, opened 2023-08-30
Trunk upgrade

InnocentBug:
# Description
Updating trunk

## Changes

there are a couple more linters.
The most potent one so far is bandit.

## Known Issues

I had to turn of bandit in the `tests` path, because it was detecting out `asserts` and our hard coded dummy passwords.
Ideally, we find a way to turn it back on.

## Checklist

- [x] My name is on the list of contributors (`CONTRIBUTORS.md`) in the pull request source branch.
- [x] I have updated the documentation to reflect my changes.

PR #301, opened 2023-08-30
La

InnocentBug:
# Description

## Checklist

### DIFF
--- a/src/cript/api/__init__.py
+++ b/src/cript/api/__init__.py
@@ -3,3 +3,6 @@
 from cript.api.api import API
 from cript.api.valid_search_modes import SearchModes
 from cript.api.vocabulary_categories import VocabCategories
+
+# Default maximum seconds for all requests to the backend
+_API_TIMEOUT = 25
--- a/src/cript/api/exceptions.py
+++ b/src/cript/api/exceptions.py
@@ -1,5 +1,5 @@
 import json
-from typing import List, Set
+from typing import List, Optional, Set
 
 from cript.exceptions import CRIPTException
 
--- a/src/cript/api/exceptions.py
+++ b/src/cript/api/exceptions.py
@@ -1,4 +1,5 @@
-from typing import List, Optional, Set
+import json
+from typing import List, Set
 
 from cript.exceptions import CRIPTException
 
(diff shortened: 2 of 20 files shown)


src/cript/api/__init__.py
```diff
@@ -3,3 +3,6 @@
 from cript.api.api import API
 from cript.api.valid_search_modes import SearchModes
 from cript.api.vocabulary_categories import VocabCategories
+
+# Default maximum seconds for all requests to the backend
+_API_TIMEOUT = 25
```
nh916, line 8: how come we are adding this?

InnocentBug: The new linter tool `bandit` is complaining if you don't have a time out.
And I think they have a good point, so adding the new version of trunk adds this.

InnocentBug:
> is this a real PR that we want merged or just to illustrate a point?

You asked to look into fixing an issue.
And this PR does that.
Merge it if you want, you don't have to.

There is a real, real PR to bump up trunks version #299

src/cript/api/__init__.py
nh916: I think the API and HTTP times out at some point anyways and I think some of the search like bigsmiles search takes a while for the response to come back which might create issues if our timeout kicks in sooner than it has time to respond.

Did we have bandit before?

InnocentBug: No Bandit is new.
We don't have a timeout set before. And that means, that connections might hang forever.

It is good practice to set one:
https://requests.readthedocs.io/en/latest/user/advanced/#timeouts

src/cript/api/exceptions.py
```diff
@@ -129,6 +130,35 @@ def __str__(self) -> str:
         return error_message
 
 
+class CRIPTDuplicateNameError(CRIPTAPISaveError):
```
nh916, line 133: is this block supposed to be in this PR?

InnocentBug: It shows up, because branch is not up to date with `develop`.
But I based the trunk update on the new `develop`.

nh916 (approved):

nh916:
I fear that merging this into both `delete` and `develop` branch may end up with merge conflicts.

I wonder if it'll be better if we merge this to `develop` and then I do a `git merge` to get the latest changes into the `delete` branch?

nh916:
I'm seeing a circular import error from our tests, I would recommend maybe creating a `api_configs.py` file and putting all configuration variables in there to avoid the import error and have it be nice and neat. We can add some comments and maybe docstring at the top too to make it even cleaner and nicer!

> Love how our tests can catch errors for us!

PR #316, opened 2023-09-02
Refactor API config variables

nh916:
# Description
Refactoring to clean up and simplify the API code.

## Changes
* Putting all configuration constant variables that will not change throughout the life time of the API inside of `cript/api/config/`
  * Then pulling those in through import statement
    * This poses a new challenge currently because the old way of overriding the variables (such as AWS S3) for tests will no longer work because the variables are coming via import statements and do not belong to api class, so they cannot be changed like normal attributes
* added `test_create_api` test back in after it was commented out

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -12,7 +12,7 @@ import jsonschema
 import requests
 from beartype import beartype
 
-from cript.api.api_config import _API_TIMEOUT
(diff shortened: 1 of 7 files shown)


nh916:
I think this might be harder to do because putting AWS config variables into another file, might be hard to overwrite in a test cleanly. We might be able to hold off on this for a while

nh916:
Brad asked not to do any more refactoring until the system is more mature

PR #317, opened 2023-09-05
fix catching the right exceptions

InnocentBug:
# Description
catching generic python exception and raising better exception for when users specify too many arguments in their constructors.

## Checklist

### DIFF
--- a/src/cript/nodes/core.py
+++ b/src/cript/nodes/core.py
@@ -82,12 +82,12 @@ class BaseNode(ABC):
                         possible_array = getattr(self._json_attrs.kwarg[: -len("_count")])
                         if not isinstance(possible_array, list):
                             raise CRIPTExtraJsonAttributes(self.node_type, kwarg)
-                    except KeyError:
+                    except AttributeError:
                         raise CRIPTExtraJsonAttributes(self.node_type, kwarg)
                 else:
                     try:
                         getattr(self._json_attrs, kwarg)
-                    except KeyError:
+                    except AttributeError:
                         raise CRIPTExtraJsonAttributes(self.node_type, kwarg)
 
         uid = get_new_uid()


nh916:
is this the code that improves `duplicate name errors` message?

nh916:
is this the code that improves `duplicate name errors` message?

InnocentBug:
Sorry, no this is the code that catches error, when users specify too many arguments in their constructors.
The `experiment` in data that Sarah had.

InnocentBug:
We should probably also uddate the documentation for this.

src/cript/nodes/core.py
```diff
@@ -82,12 +82,12 @@ def __init__(self, **kwargs):
                         possible_array = getattr(self._json_attrs.kwarg[: -len("_count")])
                         if not isinstance(possible_array, list):
                             raise CRIPTExtraJsonAttributes(self.node_type, kwarg)
-                    except KeyError:
+                    except AttributeError:
```
nh916, line 85: can this error only be raised for too many or too few arguments or do you think it can be raised other places as well?

because if it is only for too few or too many arguments then we could give an error that looks like
```
ImproperNodeCreation: 'Data' node 'my data node name' was created with too few or too many arguments
```

however, if this error can be raised for a lot of reasons and we don't know, then we can give a generic error and that should help them with debugging

nh916 (approved):

nh916:
> We should probably also uddate the documentation for this.

I would update the documentation, but not sure where to update and what to write. Feel free to create a trello ticket and tag me and I can get to it

InnocentBug:
> I would update the documentation, but not sure where to update and what to write. Feel free to create a trello ticket and tag me and I can get to it.

You already did that.
Remove Experiment from data.

nh916:
> > I would update the documentation, but not sure where to update and what to write. Feel free to create a trello ticket and tag me and I can get to it.
> 
> You already did that. Remove Experiment from data.

oh yeah got it! yeah it's already done! ✅

src/cript/nodes/core.py
nh916: okie dokes, I can do that

nh916: actually looking at the error message I think it is okay for now, but I wrote a small trello ticket with ideas for future upgrades

https://trello.com/c/M1JuhAUv

PR #321, opened 2023-09-06
API Tests: rename test & delete test

nh916:
# Description
uncommented test and changed test name to be self documenting

## Changes
* uncommented `test_create_api` to allow it to be usable again
  * not a hugely important test, but we already have it so we can use it
    * if we need to eliminate tests later to lighten up maintenance, this can be a great place to remove because the API client is test throughout the SDK anyways
* renamed functional test from `test_api_cript_env_vars` to `test_create_api_with_none`

## Checklist

### DIFF
--- a/tests/api/test_api.py
+++ b/tests/api/test_api.py
@@ -20,13 +20,11 @@ def test_create_api(cript_api: cript.API) -> None:
     """
     tests that an API object can be successfully created with host and token
     """
-    # api = cript.API(host=None, api_token=None)
-    #
-    # # assertions
-    # assert api is not None
-    # assert isinstance(api, cript.API)
+    api = cript.API(host=None, api_token=None)
 
-    pass
+    # assertions
+    assert api is not None
+    assert isinstance(api, cript.API)
 
 
 def test_api_with_invalid_host() -> None:
@@ -48,10 +46,10 @@ def test_api_context(cript_api: cript.API) -> None:
     assert cript.api.api._get_global_cached_api() is not None
 
 
-def test_api_cript_env_vars() -> None:
+def test_create_api_with_none() -> None:
     """
-    tests that when the cript.API is given None for host, api_token, storage_token that it can correctly
-    retrieve things from the env variable
+    tests that when the cript.API is given `None` for `host`, `api_token`, `storage_token`
+    that it can correctly retrieve them from the environment variables
     """
     host_value = "http://development.api.mycriptapp.org/"
     api_token_value = "my cript API token value"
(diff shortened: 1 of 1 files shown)


tests/api/test_api.py
```diff
@@ -48,10 +46,10 @@ def test_api_context(cript_api: cript.API) -> None:
...
 
-def test_api_cript_env_vars() -> None:
+def test_create_api_with_none() -> None:
     """
-    tests that when the cript.API is given None for host, api_token, storage_token that it can correctly
-    retrieve things from the env variable
+    tests that when the cript.API is given `None` for `host`, `api_token`, `storage_token`
+    that it can correctly retrieve them from the environment variables
     """
     host_value = "http://development.api.mycriptapp.org/"
```
InnocentBug, line 54: Shouldn't we have those exported in the env anyways?
When you run the tests?

It can be confusing, if you set your env variables to connect to `staging` and then you run the tests and they fail because `development` is down.

tests/api/test_api.py
```diff
@@ -20,13 +20,11 @@ def test_create_api(cript_api: cript.API) -> None:
     """
     tests that an API object can be successfully created with host and token
     """
-    # api = cript.API(host=None, api_token=None)
-    #
-    # # assertions
-    # assert api is not None
-    # assert isinstance(api, cript.API)
+    api = cript.API(host=None, api_token=None)
```
InnocentBug, line 23: This was commented because we use a fixture for this now.
We can have it, it doesn't hurt (as long as you can connect twice to cript).

But I would also be OK with deleting it entirely.

tests/api/test_api.py
nh916: yes, but I thought this would be an easier way to write the tests so its modular and not dependent on the CI variables, so we can change the CI variables without this going off that it needs to be changed too. It is more self contained in the function, easier to read, and easier to write assertions for.

I can change it if we think getting it directly from the CI is better, it is not a huge deal to me.

InnocentBug: I think we as knowledgable engineers that run the tests should be comfortable with ENV variables.
I'd like it consistent with them

tests/api/test_api.py
InnocentBug: sure,

tests/api/test_api.py
nh916: > I think we as knowledgable engineers

I appreciate the vote of confidence, but let me tell you, I've made some of the dumbest mistakes imaginable before 😂 

but yeah I can change it to depend on the environment instead of being in the function

InnocentBug (approved):

PR #326, opened 2023-09-07
downgrade `JSONSchema` package to work with Python 3.7

nh916:
# Description
downgrade `JSONSchema` package to work with Python 3.7

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 requests==2.31.0
-jsonschema==4.19.0
+jsonschema==4.17.3
 boto3==1.28.39
 beartype==0.14.1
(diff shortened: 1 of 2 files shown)


InnocentBug (approved):

InnocentBug:
I think ranges are now typically better then fixed numbers.

That gives pip more flexibility to resolve, and this probably wouldn't have happened.

nh916:
> I think ranges are now typically better then fixed numbers.
> 
> That gives pip more flexibility to resolve, and this probably wouldn't have happened.

yeah I looked into that as well and was thinking of using it, but thought what if that opens us up to more issues because each user might be on a different package and make issues harder to fix. Not sure honestly, but we can try ranges.

I think we need to for sure fix our CI to catch these issues for us because it should have been caught by CI

PR #331, opened 2023-09-11
switch `test_config_file` to use staging for `host` instead of develop

nh916:
# Description
switch `test_config_file` test to use staging host instead of develop

develop server seems to not be stable, so moving every host to staging

## Checklist


InnocentBug (approved):

nh916:
I think this is a super simple change, but I'll wait for the tests to complete before merging regardless

PR #333, opened 2023-09-11
Fix MKDocs Build Workflow Permissions

nh916:
# Description
* Giving GitHub workflow the permission to read from `main` and write to `gh-pages`
  * GitHub actions kept getting `permission denied errors` when trying to write to `gh-pages` branch, and the issue was that we did not specify that it has write permissions. 

## Checklist

### DIFF
--- a/.github/workflows/build_and_deploy_docs.yaml
+++ b/.github/workflows/build_and_deploy_docs.yaml
@@ -1,5 +1,7 @@
-# build docs from master branch and push to gh-pages branch to be deployed to repository GitHub pages
+# build docs from main branch and push to gh-pages branch to be deployed to repository GitHub pages
 name: Build & Deploy Docs
+
+# trunk-ignore(checkov/CKV2_GHA_1)
 permissions: write-all
 
 on:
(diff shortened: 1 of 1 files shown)


nh916:
after this PR we'll need to make another merge from `develop` to `main` to deploy the SDK docs on gh-pages

nh916:
trunk is giving

```
  ISSUES  

.github/workflows/build_and_deploy_docs.yaml:3:1
 3:1  high  Ensure top-level permissions are not set to write-all  checkov/CKV2_GHA_1

Checked 1 modified file
✖ 1 new issue
✖ Failed to post annotations to github: GitHub Error:403 Body:{"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/checks/runs#create-a-check-run"}
Error: Process completed with exit code 1.
```

I think it is complaining that `write-all` is bad and the permissions should be much more granular. I agree with this and wanted to do the same thing, but I don't think there is a way in GitHub workflow files to specify `read from main and write to gh-pages` so I think it will have to remain as `write-all`

@InnocentBug if you have any better ideas feel free to let me know or push to the branch directly

InnocentBug (approved):

PR #341, opened 2023-09-12
SDK Requires Python 3.8+

nh916:
# Description
updated SDK to require Python 3.8+ because [Python 3.7 is deprecated](https://devguide.python.org/versions/). Dropping support for SDK 3.7

## Changes
* changed README Python badge to be `3.8+`
* changed README installation guide to be `3.8+`
* changed `tests.yaml` to run from `[3.8, 3.11]`

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
 
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.7, 3.11]
+        python-version: [3.8, 3.11]
 
     env:
       CRIPT_HOST: https://lb-stage.mycriptapp.org/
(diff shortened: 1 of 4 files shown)


nh916:
> I am not a fan of this. Yes, Python 3.7 is not a good version anymore, but we may want to support legacy if it doesn't break anything from our side. We don't want to lose users because of this.
> 
> Plus, it can be very difficult for some users to update their environment. (HPC clusters etc). So, if it it doesn't break anything, I would continue to support 3.7. If it starts breaking things, we can reconsider.

I hear what you're saying, and I'll hold off for a bit and it should be fine.

I would push us to migrate to Python 3.8+ quickly though because we do not yet have a massive pool of users using our system so we can afford to do breaking changes and upgrades pretty quickly and painlessly. If we support outdated versions for too long then we might start having users that are stuck on Python 3.7 and then need to support them, which could become a pain overall.

But for now I think we can leave it as Python 3.7+ and that should be fine

nh916:
marking this as draft, so we can merge it when we need to later

nh916:
seeing how all the packages that we are using are dropping support for Python 3.7, I strongly [recommend we do the same](https://github.com/C-Accel-CRIPT/Python-SDK/pull/341), we don't currently have anyone that requires Python 3.7 for their work

InnocentBug (approved):

PR #344, opened 2023-09-13
renaming CI names to be more self documenting

nh916:
# Description
was setting up status check repository requirements and saw that some of our names were not very self documenting, so changing that here quickly.

I debated if we should use dashes or underscores for job names, and ultimately I came up with underscores because if we need to chain jobs together calling the job would be easier I think and we are using underscore convention everywhere in our repository so it just makes sense to keep the consistency. Feel free to let me know if I am wrong

## Checklist

### DIFF
--- a/.github/workflows/build_and_deploy_docs.yaml
+++ b/.github/workflows/build_and_deploy_docs.yaml
@@ -13,7 +13,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  deploy:
+  deploy_docs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


InnocentBug (commented):
At present we keep those names also in .trunk/trunk.yaml for the merge-q. 
hntil we remove that. We need to keep it in. Sync with new names.

nh916:
> At present we keep those names also in .trunk/trunk.yaml for the merge-q. hntil we remove that. We need to keep it in. Sync with new names.

okay I did not know that. In that case, I'll go ahead and put the required checks to be based off of the current CI and we'll change it after we rename them and replace the merge queue

InnocentBug:
not prio right now

PR #350, opened 2023-09-15
Upgrade `requirements_docs.txt`

nh916:
# Description
Upgrade requirements_docs.txt

## Notes
requirements_dev.txt needed upgrades, but not sure if we want to change our syntax from `>=` to `==`
image

---

* mkdocs 9.3 dropped support for Python 3.7
* pymdown-extensions 10.3 dropped support for Python 3.7

## Checklist

### DIFF
--- a/requirements_docs.txt
+++ b/requirements_docs.txt
@@ -1,5 +1,5 @@
 mkdocs==1.5.2
-mkdocs-material==9.2.6
-mkdocstrings[python]==0.22.0
-pymdown-extensions==10.2.1
+mkdocs-material==9.3.1
+mkdocstrings[python]==0.23.0
+pymdown-extensions==10.3
 jupytext==1.15.1


nh916:
seeing how all the packages that we are using are dropping support for Python 3.7, I strongly [recommend we do the same](https://github.com/C-Accel-CRIPT/Python-SDK/pull/341), we don't currently have anyone that requires Python 3.7 for their work

InnocentBug:
> we don't currently have anyone that requires Python 3.7 for their work

we don't know our users, so how can we know?

nh916:
> > we don't currently have anyone that requires Python 3.7 for their work
> 
> we don't know our users, so how can we know?

Currently, Sarah appears to be the sole user of the SDK, and she has transitioned to Python 3.8+. Users on Python 3.7 would probably upgrade given that many other packages now require 3.8+. If they have any concerns, they can reach out to us via our discussions or other means. We'll either assist them directly or we'll reconsider certain updates based on their feedback.

InnocentBug (approved):

PR #351, opened 2023-09-15
refactoring vocab to its own file

nh916:
# Description
This PR puts everything that has to do with vocab into its own module, but from the user perspective they can still use it like `cript.API.get_vocab()` and it will work correctly.

## Checklist


nh916:
Brad said to not do any refactoring for now until later when the system is more mature

PR #359, opened 2023-09-19
upgraded `api.delete_node_by_uuid()` to use str or node class type

nh916:
# Description
upgraded `cript.API.delete_node_by_uuid(node_type, uuid)` to use str or node for `node_type`

## Changes
* can pass in either class name of node_type or string
  * updated documentation to reflect changes
  * updated code examples for both versions
* created utility function `get_node_type_snake_case`
  * allows you to get the node_type from class, object, or string
  * wrote a test for it as well
    * test is passing fine

### Screenshots
delete node by uuid updated docs screenshot

## Notes
* did not write any integration tests for `delete_node_by_uuid` testing snake case string node type, class, or object.
  * relying on the integration tests for each node, since each node calls `delete` and that calls `delete_node_by_uuid`

## Checklist


nh916:
@InnocentBug tried formatting `api.py` but it keeps having issues, can you give it a try when you get a chance

src/cript/api/utils/helper_functions.py
```diff
@@ -41,3 +41,30 @@ def _get_node_type_from_json(node_json: Union[Dict, str]) -> str:
     # if invalid then raise error
     else:
         raise CRIPTJsonNodeError(node_list=node_type_list, json_str=str(node_json))
+
+
+def get_node_type_snake_case(node_type: Union[Any, str]) -> str:
```
InnocentBug, line 46: This is a duplication of  `BaseNode.node_type_snake_case`.

tests/api/test_api_utils.py
```diff
@@ -0,0 +1,17 @@
+import cript
+from cript.api.utils.helper_functions import get_node_type_snake_case
+
+
+def test_get_node_type_snake_case(simple_computation_process_node) -> None:
```
InnocentBug, line 5: the test is good, but it should test the BaseNode function

src/cript/api/api.py
```diff
@@ -1241,11 +1233,13 @@ def delete_node_by_uuid(self, node_type: str, node_uuid: str) -> None:
...
+
+        delete_node_api_url: str = f"{self._host}/{node_type_snake_case}/{node_uuid}/"
 
         response: Dict = requests.delete(headers=self._http_headers, url=delete_node_api_url, timeout=_API_TIMEOUT).json()
 
         if response["code"] != 200:
             raise APIError(api_error=str(response), http_method="DELETE", api_url=delete_node_api_url)
 
-        self.logger.info(f"Deleted '{node_type.title()}' with UUID of '{node_uuid}' from CRIPT API.")
+        self.logger.info(f"Deleted '{node_type_snake_case.title()}' with UUID of '{node_uuid}' from CRIPT API.")
```
InnocentBug, line 1245: use the basenode functions

src/cript/api/api.py
```diff
@@ -1172,35 +1159,40 @@ def delete(self, node) -> None:
         self.delete_node_by_uuid(node_type=node.node_type_snake_case, node_uuid=str(node.uuid))
 
     @beartype
-    def delete_node_by_uuid(self, node_type: str, node_uuid: str) -> None:
+    def delete_node_by_uuid(self, node_type: Union[Any, str], node_uuid: str) -> None:
```
InnocentBug, line 1162: ideally, you don't have to specify the node type at all.
Check out the class attribute `UUIDBaseNode._uuid_cache`. 
It is a dictionary that maps UUID to all nodes that have been encountered with the current SDK session.

Use that to look up the node by uuid.

Only case this doesn't work, would be if the node is only present in the backend, in that case a simple `GET` to the back end should do the trick.

InnocentBug (changes requested):

src/cript/api/utils/helper_functions.py
nh916: they are different because that has access to the node class and is just taking it and converting it to string, allows the user to specify either a class if they have it or a string if they need to and this takes both and gives back the same str answer regardless.

src/cript/api/api.py
nh916: that is if I have access to the class, but if I have access to only the string or class then I'll have to write the if logic at the beginning of the file and always get the str name from the class/object

src/cript/api/api.py
nh916: that I did not think about. I'm not sure if that should go into this method or the method that is calling it. We could also have a utility method that uses this method as a wrapper that both looks up the node by uuid and then deletes it.

I really like this delete because I think that is really useful, and this method is really clean and takes only what is needed and deletes it.

if we want to have delete method that both looks up the UUID and then performs a delete, then I'd be in favor of making another method that does that, but would prefer if the user passes in their arguments instead so it is no room for ambiguity.

InnocentBug: The reason I am not a big fan of this delete is, that a) it introduces code duplication with the snake case and b) it puts burden on the user (providing the node type).
And we can eliminate this burden and should do that, because i) it is easier on the user and ii) if we don't the user can for example give you a wrong node type and debugging this is annoying for a user.

nh916: I hear that, I'm kind of thinking that since I already wrote it, it is clean, and could be useful to keep it and provide a new method called `find_and_delete_node_by_uuid` that finds the node by uuid and deletes it. 

I'm not sure if it'll create code duplication with multiple delete methods or give more options. I have a feeling that this method will end up being inevitable at the end because other methods will just be a wrapper around this, but honestly not sure. What are your thoughts?

PR #360, opened 2023-09-20
refactored `prepare_host()` into a function to reduce cript.API file bulk

nh916:
# Description
was thinking how I can make the `cript.API` file more manageable and smaller chunks, and saw that the `prepare_host()` can easily be a function in its own file as a reference, so I quickly did that

## Changes
* easily refactored and put `prepare_host()` into a function that reduces the bulk of the api.py file
* wrote docstrings for `prepare_host()`
* optimized imports after moving `prepare_host()` to its own function in another file
* formatted everything with black

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -3,7 +3,6 @@ import json
 import logging
 import os
 import uuid
-import warnings
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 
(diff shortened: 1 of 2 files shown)


nh916:
> I am a big fan of this change. It is merely addressing the symptom of a too big file.
> 
> Working with the code will actually be more stressfull and annoying, because now you have to look up too different source files.
> 
> The real solution to this issue is the refactor API into multiple sub classes compositioned into one API class.
> 
> Like DBSchema, FileSaver, NodeSaver etc. Also much better for unit tests.

I think this might be something that we'll need a experienced software engineer outside perspective on because I think we both have very strong opinions on this. From my experience, I can just ctrl click on a function and jump to it and see how it operates and jump back with ease or just hover my mouse over it and see the docstrings to see its purpose and move on, and only work through the implementation when needed. I think multiple small functions are a lot easier to maintain than multiple classes and easier to work with too.

I see 2 potential options, 
1. leave this until later because it is not high priority right now
2. get an outside perspective

I think I'm leaning towards option 1 because option 2 would probably create a lot of back and forth and would slow the SDK down and I would rather not debate over them right now.

nh916:
closing this PR now, but I am happy to open it if we want to go further with it

PR #364, opened 2023-09-22
added `has_db_schema_validation` flag

nh916:
# Description
I added a flag to the entire class that can enable or disable the db schema validation

## Notes
@InnocentBug not sure if we need the flag added to the whole class or if we can just put it on the method instead as a parameter. 

I think passing it as an argument is more preferable because it avoids the bulk, and encapsulates everything just in the method.

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -602,7 +602,7 @@ class API:
             return self._db_schema
 
     @beartype
-    def _is_node_schema_valid(self, node_json: str, is_patch: bool = False) -> bool:
+    def _is_node_schema_valid(self, node_json: str, is_patch: bool = False, force_validation: bool = False) -> bool:
         """
         Checks a node JSON schema against the db schema to return if it is valid or not.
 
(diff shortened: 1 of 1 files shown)


nh916:
I don't love this implementation because it is either validating everything or nothing, but I think this is just a stepping stone for us to then use it in the next parts of the SDK and allow for turning OFF of some db schema validations and leave the essential ones ON

src/cript/api/api.py
```diff
@@ -603,7 +604,7 @@ def _get_db_schema(self) -> dict:
     @beartype
     def _is_node_schema_valid(self, node_json: str, is_patch: bool = False) -> bool:
```
InnocentBug, line 605: I would write a `force_schema_validation` argument into this function.
This way, there is a way to force the validation.

And the we can force this validation in `api._internal_save`.

That looks like the best solution to me.

InnocentBug (changes requested):

src/cript/api/api.py
nh916: added

PR #365, opened 2023-09-22
Paginator get node by name

nh916:
# Description
wrote a method within paginator called `get_node_by_name` where the user can get the node they want by specifying the name instead of the index number like before. I think this could help with UX

## Changes
* wrote `get_node_by_name` in cript.Paginator
* wrote docstrings for `get_node_by_name`
* wrote test for `get_node_by_name` and it works fine

### Screenshots
get_node_by_name from pagiantor docs screenshot

## Notes
* I wrote this method to get the node by name from the current page
  * not sure if we want to have the user specify the name and get it from the current page or if we want to have the user specify the name, and this method will flip through as many pages as needed to get the node or return None

## Checklist


src/cript/api/paginator.py
```diff
@@ -230,3 +230,50 @@ def fetch_page_from_api(self) -> List[dict]:
...
+        Returns
+        -------
+        Optional[Dict]
+            the desired node in JSON form and if it doesn't exist then it returns `None`
+        """
+        # return find_json_node_by_name(list_of_nodes=self.current_page_results, target_name=node_name)
+
+        for node_json in self.current_page_results:
+            if node_name == node_json["name"]:
+                return node_json
```
InnocentBug, line 277: I would prefer if it returns a node, not a JSON dict.

nh916: we can do that, but everywhere else that paginator is working it is returning a JSON and not a node, we would have to change it everywhere and the `api.get_nodby_uuid()` would then be useless and need changing

InnocentBug: hmm, not sure what to do with this

PR #369, opened 2023-09-26
added a flag for if the file is a local path or not

nh916:
# Description
added a flag for if the file is a local path or not

## Changes

* changed property setter to a method setter
  * setter takes `new_source` and a boolean flag of if the file is local or not
* changed constructor to require boolean flag of `is_file_source_local_path`
* wrote `set_file_source()` docstring sna documentation
* updated source property docstrings documentation to reflect the new changes
* changed `_upload_file_and_get_object_name` to work with boolean flag
  * removed redundant check on it 
* changed is statement check on `ensure_uploaded()` to work with boolean flag
* fixed supporting_nodes.py file fixture to work with new boolean flag `is_file_source_local_path`
* fixed test_file.py to pass with new boolean flag `is_file_source_local_path`

## Checklist

### DIFF
--- a/src/cript/nodes/supporting_nodes/file.py
+++ b/src/cript/nodes/supporting_nodes/file.py
@@ -64,15 +64,14 @@ def _upload_file_and_get_object_name(source: Union[str, Path], api=None) -> str:
     """
     from cript.api.api import _get_global_cached_api
 
-    # convert source to str for `_is_local_file` and to return str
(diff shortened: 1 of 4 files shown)


nh916:
@InnocentBug can't add boolean flag for `is_file_source_local_path` because then serialization/deserialization will break because it will not have that variable within the JSON when deserializing

PR #370, opened 2023-09-26
Require file extension for `cript.File` node

nh916:
# Description

## Changes
* requiring file extension for all nodes when created
* updated documentation
  * updated File node creation example code
    * changed the code examples from minimal and maximal to 2 maximal file nodes with local path source and web url source
  * fixed `ensure_uploaded` documentation
    * the syntax was wrong
  * fixed example code documentation after the error popped up
    > **really happy that there is automated tests that check the validity of the code** 
    * formatted entire example code walkthrough and comments

### Screenshots

<details>

<summary> click to see updated file documentation </summary>

image

</details>

<details>

<summary> click to see formatted code example documentation </summary>

image

</details>

## Checklist

### DIFF
--- a/docs/examples/simulation.md
+++ b/docs/examples/simulation.md
@@ -250,10 +250,33 @@ experiment.computation += [init, equilibration, bulk, ana]
 New we'd like to upload files associated with our simulation. First, we'll instantiate our File nodes under a specific project.
 
 ```python
-packing_file = cript.File(name="Initial simulation box snapshot with roughly packed molecules", type="computation_snapshot", source="path/to/local/file", extension=".csv")
-forcefield_file = cript.File(name="Forcefield definition file", type="data", source="path/to/local/file", extension=".pdf")
-snap_file = cript.File(name="Bulk measurement initial system snap shot", type="computation_snapshot", source="path/to/local/file", extension=".png")
-final_file = cript.File(name="Final snapshot of the system at the end the simulations", type="computation_snapshot", source="path/to/local/file", extension=".jpeg")
+packing_file = cript.File(
+    name="Initial simulation box snapshot with roughly packed molecules",
+    type="computation_snapshot",
+    source="path/to/local/file",
+    extension=".csv",
+)
+
+forcefield_file = cript.File(
+    name="Forcefield definition file",
+    type="data",
+    source="path/to/local/file",
+    extension=".pdf",
+)
+
+snap_file = cript.File(
+    name="Bulk measurement initial system snap shot",
+    type="computation_snapshot",
+    source="path/to/local/file",
+    extension=".png",
+)
+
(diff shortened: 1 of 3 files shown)


nh916:
do we want to technically allow users to set extension to empty string or block any attempts?

nh916:
not sure if we should speak with the entire team on the file extension formats, but I just went with what was described in the [data model paper](https://pubs.acs.org/doi/suppl/10.1021/acscentsci.3c00011/suppl_file/oc3c00011_si_001.pdf#page=28)

InnocentBug:
Well, wouldn't we ideally figure out the extension from the filename, and only let the users overwrite the explict extension if necessary?

Also this change has a lot of updated white space changes, that I don't really think look great.

nh916:
> Well, wouldn't we ideally figure out the extension from the filename, and only let the users overwrite the explict extension if necessary?

we could do that, but for v1 I was thinking of having it this way and then for the next PR I can improve on it where I can figure it out. Should I put it into this PR?

InnocentBug (approved):

nh916:
> Also this change has a lot of updated white space changes, that I don't really think look great.

The new spaces makes it look a million times better and easier to read. I find that scrolling through code is typically makes it harder for me to read it.

I should have put the formatting in a different PR, but I just didn't. I could try to break it up into a different PR though

InnocentBug:
We just need to be a little careful, we don't want to break code if possible.
And this changes our API.

nh916:
@InnocentBug are you still okay with me merging this after my new comments/new information?

nh916:
> We just need to be a little careful, we don't want to break code if possible. And this changes our API.

I don't understand, tell me more

InnocentBug:
> I don't understand, tell me more

Sarah wrote code with the SDK that will break after this merge.
She didn't specify the extension for her files.

nh916:
> > I don't understand, tell me more
> 
> Sarah wrote code with the SDK that will break after this merge. She didn't specify the extension for her files.

oh true, hmmm, I can think of 2 ways of going about this. 
1. not include this PR
2. next version we make it into v3.0.0, since there is a breaking change and we let users know how to migrate to the new version of the SDK?


we could not include this since the data model has this as optional, but I recommend having it because otherwise the file might not be read correctly. I think OS have their own magic that they can mostly correctly guess the file type, but not always so this would make it a bit easier and make the file node more complete in my own opinion. 
> However, it would be a discrepancy between data model and SDK.

I'm thinking in future SDK release we put that this is a breaking change and how to migrate to the new version.

what are your thoughts?

InnocentBug:
I still think we can merge, but ideally we have the feature in that figures out the extension from the filename before the next release.
That way no code will be broken.

PR #374, opened 2023-09-29
Performance improvement for node validation

InnocentBug:
# Description
Addition or replacement of #361 
This accelerates the validation of nodes.

When we change a given node A possibly by adding node B to be its child, we already know that A that B is a valid node.

So, instead of validating the entire graph of A, it would be more efficient to only validate A and not its entire graph of children.
So the idea is to reduce the down stream graph of A to only UUID.

However, there is a flaw with this B might be a valid node, but it might be in an invalid attribute of A.
The data schema is flawed that way, that it would accept this bad graph. (The data schema has no way of validating the node type if give a UUID only.)
We should probably write a ticket for this flaw too.
It could be averted if `UID` and `UUID` also require `node` in the JSON.

As a solution, we express A and all its direct children directly, that makes sure that A is still correct and all its children are also of correct type, but still saves on evaluating the entire downstream graph.

## Changes

Added a new mode of encoding nodes to JSON, where only a fixed list of UUID are fully expressed and every other node is just a UUID.

## Tests

The existing tests should cover this.

## Known Issues

See above, flaw in the data schema that doesn't evaluate node types of UUID and UID.
Not new with the PR, but found writing this PR.

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -713,7 +713,7 @@ class API:
         for file_node in node.find_children({"node": ["File"]}):
             file_node.ensure_uploaded(api=self)
 
-        node.validate()
+        node.validate(api=self, only_this_node_full=False)
 
         # Dummy response to have a virtual do-while loop, instead of while loop.
         response = {"code": -1}
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -713,6 +713,7 @@ class API:
         for file_node in node.find_children({"node": ["File"]}):
             file_node.ensure_uploaded(api=self)
 
+        # We request a full validation of the node just to be sure before save
         node.validate(api=self, only_this_node_full=False)
 
         # Dummy response to have a virtual do-while loop, instead of while loop.
--- a/src/cript/nodes/core.py
+++ b/src/cript/nodes/core.py
@@ -142,7 +142,7 @@ class BaseNode(ABC):
             self._json_attrs = old_json_attrs
             raise exc
 
-    def validate(self, api=None, is_patch=False) -> None:
+    def validate(self, api=None, is_patch=False, only_this_node_full=True) -> None:
         """
         Validate this node (and all its children) against the schema provided by the data bank.
 
@@ -154,7 +154,10 @@ class BaseNode(ABC):
 
         if api is None:
             api = _get_global_cached_api()
-        api._is_node_schema_valid(self.get_json(is_patch=is_patch).json, is_patch=is_patch)
+        # We request only this node and all its **direct** children full, everything else is UUID.
+        # As a result, we avoid evaluating the entire graph, and only validate relevant parts
+        json_str: str = self.get_json(is_patch=is_patch, only_this_node_full=only_this_node_full).json
+        api._is_node_schema_valid(json_str, is_patch=is_patch)
 
     @classmethod
     def _from_json(cls, json_dict: dict):
@@ -248,6 +251,7 @@ class BaseNode(ABC):
         known_uuid: Optional[Set[str]] = None,
         suppress_attributes: Optional[Dict[str, Set[str]]] = None,
         is_patch: bool = False,
+        only_this_node_full=False,
         condense_to_uuid: Dict[str, Set[str]] = {
             "Material": {"parent_material", "component"},
             "Experiment": {"data"},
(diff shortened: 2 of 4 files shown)


tests/test_node_util.py
```diff
@@ -142,10 +141,6 @@ def test_json_error(complex_parameter_node):
     parameter._json_attrs = replace(parameter._json_attrs, value="abc")
     with pytest.raises(CRIPTNodeSchemaError):
         parameter.validate()
-    # Let's break it completely
-    parameter._json_attrs = None
```
InnocentBug, line 146: I removed this test.
It was pretty artificial and didn't work with the new way anymore.

InnocentBug:
Locally, this reduced the time to run `pytest -s test/test_node_util.py` from 3:30 min to 1:30.
So, pretty good improvement.

nh916:
> Locally, this reduced the time to run `pytest -s test/test_node_util.py` from 3:30 min to 1:30. So, pretty good improvement.

were you able to run all the tests and get the wait time before and after the implementation?

src/cript/nodes/core.py
```diff
@@ -296,6 +300,26 @@ class ReturnTuple:
         previous_condense_to_uuid = copy.deepcopy(NodeEncoder.condense_to_uuid)
         NodeEncoder.condense_to_uuid = condense_to_uuid
 
+        # For validation of just a node and its direct children, we enable a new mode in Node encoding
+        # For this node we explicitly list the nodes that are fully expressed, every other node is a UUID
+        previous_only_not_uuid: Optional[list[str]] = NodeEncoder.only_not_uuid
```
nh916, line 305: is it possible to make the variable names more self documenting?

src/cript/nodes/core.py
```diff
@@ -296,6 +300,26 @@ class ReturnTuple:
...
+        if only_this_node_full:
```
nh916, line 306: if this is to separate just the direct child nodes, then maybe it should be in its own function/method and gets called from here, goes in there, separates out what it needs, and returns the data to here

src/cript/nodes/core.py
```diff
@@ -308,6 +332,7 @@ class ReturnTuple:
             NodeEncoder.known_uuid = previous_known_uuid
             NodeEncoder.suppress_attributes = previous_suppress_attributes
             NodeEncoder.condense_to_uuid = previous_condense_to_uuid
+            NodeEncoder.only_not_uuid = previous_only_not_uuid
```
nh916, line 335: more self documenting variable names if possible

nh916:
I think something that could be a good idea is maybe we figure out where the validation is essential, and where it is not essential because something else is already enforcing validation anyways, and we could try to put the validations only in the essential areas and see how that works, if it is too little validation then we add more, we keep iterating until we get it perfect so it is not too much or too little.

src/cript/api/api.py
```diff
@@ -713,7 +713,8 @@ def _internal_save(self, node, save_values: Optional[_InternalSaveValues] = None
         for file_node in node.find_children({"node": ["File"]}):
             file_node.ensure_uploaded(api=self)
 
-        node.validate()
+        # We request a full validation of the node just to be sure before save
+        node.validate(api=self, only_this_node_full=False)
```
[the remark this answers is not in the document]
InnocentBug: I honestly don't know what you mean by better variable name.

calling `validate` with `only this node full` is almost a complete English sentence.
And it says what it is.
We validate only this node in full.

src/cript/nodes/core.py
InnocentBug: Not sure, what do you have in mind?

src/cript/nodes/core.py
InnocentBug: no, I don't think that is a good idea.
It makes the logic convulted and complicated.

src/cript/nodes/core.py
InnocentBug: see above/

src/cript/api/api.py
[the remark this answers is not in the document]
nh916: fair point!

src/cript/nodes/core.py
nh916: I think it works, if we can refactor later if needed

tests/test_node_util.py
nh916: just be sure that the code you are putting in will not break anything and it is HEAVILY tested if possible

nh916 (approved):

PR #397, opened 2023-10-19
node name min length checker

nh916:
# Description

## Issue Link
* https://trello.com/c/NeIiLogy

## Checklist


nh916:
we can do this check ourselves, however, it should be done by the API via the DB schema. I think the issue is that the name field is currently set to a min of 2 when it should be a min of 3

PR #398, opened 2023-10-24
`load_nodes_from_json` works with both str and dict args

nh916:
# Description
`load_nodes_from_json` works with both str and dict args
this makes the function a more versatile and makes user scripts easier to write since they don't have to change formats.

The code written for it is pretty clean and nice to work with I think.

## Changes
* `load_nodes_from_json` accepts with JSON str or dict arguments
* wrote a small test for `load_nodes_from_json` using a JSON dict argument

## Checklist


nh916:
need to update code example with doctest after this PR is in https://github.com/C-Accel-CRIPT/Python-SDK/pull/371

InnocentBug (approved):

PR #409, opened 2023-11-13
removed `get_node_by_uuid` cript.API feature

nh916:
# Description
removed `get_node_by_uuid` cript.API feature. This method will be removed in later versions in favor of the single search `cript.API.get_node_by_exact_match()`

I thought it'd be a better idea to remove it from the SDK and not release it in this version because if users use it, and we remove it in the next version it will introduce a breaking change that will break their script and it will be a hassle, better to not introduce at all this version and work on it in the next version.

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -1143,44 +1143,6 @@ class API:
 
         return Paginator(http_headers=self._http_headers, api_endpoint=api_endpoint, query=value_to_search, current_page_number=page_number)
 
-    @beartype
-    def get_node_by_uuid(self, node_type: Any, node_uuid: str) -> Any:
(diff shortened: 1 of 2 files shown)


nh916:
@InnocentBug if you can please see if you can format it with trunk as I am running `trunk fmt` on my system through WSL, but it is not formatting anything or finding any issues

InnocentBug:
done

InnocentBug (approved):

nh916:
> done

You are the best! ✅ 🚀

nh916:
it's a shame this had to go because I worked hard on it too, but the new feature will be even better!

PR #411, opened 2023-11-13
Skip validation

InnocentBug:
# Description
As discussed with @bearmit 
A flag that disables partially the tests is being implemented.

At the moment we do not disable tests by default.

There is a text output that let's users know that they can disable the validation if needed.
Some checks are forced even as tests are disabled.
 - upon `save`
 - and upon `node.json`

## Changes

I changed the logging  text to let users know that they can disable the validation.
And marked forced checks as such, letting the user know that they can't be disabled and that if they have issues, they should enable the tests.

## Tests
I added a test that checks, that a wrong node is accepted if tests are disabled, but detected if forced.

## Known Issues
None.
## Notes

If we decide to disable tests by default, we just have to change the default value for `cript.API.skip_validation` to `True`.

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -615,7 +615,7 @@ class API:
             return self._db_schema
 
     @beartype
-    def _is_node_schema_valid(self, node_json: str, is_patch: bool = False, force_validation: bool = False) -> bool:
+    def _is_node_schema_valid(self, node_json: str, is_patch: bool = False, force_validation: bool = False) -> Union[bool, None]:
         """
         checks a node JSON schema against the db schema to return if it is valid or not.
 
@@ -651,7 +651,7 @@ class API:
 
         # Fast exit without validation
         if self.skip_validation and not force_validation:
-            return True
+            return None
 
         db_schema = self._get_db_schema()
 
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -81,6 +81,12 @@ class API:
     _internal_s3_client: Any = None  # type: ignore
     # trunk-ignore-end(cspell)
 
+    # Advanced User Tip: Disabling Node Validation
+    # For experienced users, deactivating node validation during creation can be a time-saver.
+    # Note that the complete node graph will still undergo validation before being saved to the back end.
+    # Caution: It's advisable to keep validation active while debugging scripts, as disabling it can delay error notifications and complicate the debugging process.
+    skip_validation: bool = False
+
     @beartype
     def __init__(self, host: Union[str, None] = None, api_token: Union[str, None] = None, storage_token: Union[str, None] = None, config_file_path: Union[str, Path] = ""):
         """
@@ -609,7 +615,7 @@ class API:
             return self._db_schema
 
     @beartype
-    def _is_node_schema_valid(self, node_json: str, is_patch: bool = False) -> bool:
+    def _is_node_schema_valid(self, node_json: str, is_patch: bool = False, force_validation: bool = False) -> bool:
         """
         checks a node JSON schema against the db schema to return if it is valid or not.
 
@@ -643,6 +649,10 @@ class API:
             whether the node JSON is valid or not
         """
 
+        # Fast exit without validation
+        if self.skip_validation and not force_validation:
+            return True
+
         db_schema = self._get_db_schema()
 
         node_type: str = _get_node_type_from_json(node_json=node_json)
@@ -651,7 +661,13 @@ class API:
 
         # logging out info to the terminal for the user feedback
         # (improve UX because the program is currently slow)
-        self.logger.info(f"Validating {node_type} graph...")
+        log_message = f"Validating {node_type} graph..."
+        if force_validation:
+            log_message = "Forced: " + log_message + " if error occur, try setting `cript.API.skip_validation = False` for debugging."
+        else:
+            log_message += " (Can be disabled by setting `cript.API.skip_validation = True`.)"
+
+        self.logger.info(log_message)
 
         # set the schema to test against http POST or PATCH of DB Schema
         schema_http_method: str
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -737,7 +737,7 @@ class API:
         for file_node in node.find_children({"node": ["File"]}):
             file_node.ensure_uploaded(api=self)
 
-        node.validate()
+        node.validate(force_validation=True)
 
         # Dummy response to have a virtual do-while loop, instead of while loop.
         response = {"code": -1}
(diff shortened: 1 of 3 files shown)


src/cript/api/api.py
```diff
@@ -81,6 +81,12 @@ class API:
     _internal_s3_client: Any = None  # type: ignore
     # trunk-ignore-end(cspell)
 
+    # Advanced User Tip: Disabling Node Validation
```
InnocentBug, line 84: Can we make sure this shows up in the documentation of this attribute?

nh916: as far as I know MkDocs will hide all private attributes, but I think we can put it in the class `attributes` docstrings and it will show up in the documentation. Am I understanding you correctly?

nh916:
please be sure this targets the `develop` branch as well

src/cript/api/api.py
InnocentBug: yes, let's put it there.
It is not a private attribute though (no `_`).

Do you know how the format looks like for that?

nh916 (dismissed):

nh916:
> > please be sure this targets the `develop` branch as well
> 
> yeah, done for both. (and branches rebased)

You are the best!

src/cript/api/api.py
nh916: yep!

this is not the PR that I did it on, but this is the docstrings format for it
https://github.com/C-Accel-CRIPT/Python-SDK/blob/8e50ca1f87a8ab9b2291d13de86ce77142955d01/src/cript/api/api.py#L54-L75

I can also help as well if you bump into issues as I have a lot of experience with documentation

nh916: if it is not a private variable and not needed for internal, then we can probably rename/refactor that part as well, I think that would be even cleaner, but it would probably need more work and probably a different PR.

tests/api/test_api.py
```diff
@@ -168,15 +168,16 @@ def test_is_node_schema_valid_skipped(cript_api: cript.API) -> None:
 
     """
 
-    cript_api.skip_validation = True
-    # ------ invalid node schema------
-    invalid_schema = {"invalid key": "invalid value", "node": ["Material"]}
+    with cript.API(host=cript_api.host, api_token=cript_api._api_token, storage_token=cript_api._storage_token) as local_cript_api:
+        local_cript_api.skip_validation = True
```
nh916, line 172: I don't love this because we are repeating code and it looks a bit weird, but we can add this as a tech debt into our trello and refactor and upgrade this after this release, if that works with you

InnocentBug: I see.
It is necessary to use a local API instance here. Otherwise we are changing the API for every other test as well.
There is a way to use `function scope` on the pytest fixture that would work too.

src/cript/api/api.py
```diff
@@ -609,7 +615,7 @@ def _get_db_schema(self) -> dict:
             return self._db_schema
 
     @beartype
-    def _is_node_schema_valid(self, node_json: str, is_patch: bool = False) -> bool:
+    def _is_node_schema_valid(self, node_json: str, is_patch: bool = False, force_validation: bool = False) -> bool:
```
nh916, line 618: I would have called it `force_schema_validation` just in case we have multiple validations in the future, but this works as well

src/cript/api/api.py
```diff
@@ -643,6 +649,10 @@ def _is_node_schema_valid(self, node_json: str, is_patch: bool = False) -> bool:
             whether the node JSON is valid or not
         """
 
+        # Fast exit without validation
+        if self.skip_validation and not force_validation:
+            return True
```
nh916, line 654: should we just return if we are not validating?
I think this version works, but I wonder if that is more accurate or not, not sure

nh916 (approved):

src/cript/api/api.py
InnocentBug: fair point, but I think we can refactor if need be.
Actually there other validations we still run at the moment.

src/cript/api/api.py
InnocentBug: we specify that we return a bool with type hints, so it has to be True.

nh916: should we change it to returns boolean or none?
I'm thinking that way it doesn't mislead the user if they forget to turn it ON/OFF or something. what do you think?

InnocentBug: I see what you mean, but I don't think it matters much.
I can implement that.

InnocentBug: done

tests/api/test_api.py
```diff
@@ -155,6 +155,40 @@ def test_is_node_schema_valid(cript_api: cript.API) -> None:
...
+        # Split the URL by "//" first to separate the scheme (like http, https)
+        parts = url.split("//", 1)
+        scheme, rest = parts if len(parts) > 1 else ("", parts[0])
+
+        # Split the rest by the first "/" to separate the domain
+        domain = rest.split("/", 1)[0]
+        return f"{scheme}//{domain}" if scheme else domain
+
+    with cript.API(host=extract_base_url(cript_api.host), api_token=cript_api._api_token, storage_token=cript_api._storage_token) as local_cript_api:
+        local_cript_api.skip_validation = True
```
nh916, line 181: will this not work with `os.getenv("CRIPT_HOST")` ?

this is a bit messy and if we can avoid it all the better, but if necessary then we can leave it in

nh916:
I want to approve this but CI is failing

tests/api/test_api.py
InnocentBug: Using the environment variable here would make this brittle.
We are offering more ways then just ENV variables to specify the host, so relying on the fact that is in the env variables for this test is not a good idea.

The ideal solution is, that we offer a deepcopy implementation for API.

The second best solution is what I did here.
I can see that having this host separation function here isn't great.
But it is only in the tests.
we could also fix our `host` property of cript.API to actually only return the host address without the `api/v2` thing at the end.

InnocentBug:
> I want to approve this but CI is failing

should be fixed.

tests/api/test_api.py
```diff
@@ -183,7 +183,7 @@ def extract_base_url(url):
         invalid_schema = {"invalid key": "invalid value", "node": ["Material"]}
 
         # Test should be skipped
-        assert local_cript_api._is_node_schema_valid(node_json=json.dumps(invalid_schema), is_patch=False) is True
+        assert local_cript_api._is_node_schema_valid(node_json=json.dumps(invalid_schema), is_patch=False) is None
```
nh916, line 186: Beautiful!

nh916 (approved):

PR #412, opened 2023-11-13
Self-Contained JSON file format

InnocentBug:
# Description

The SDK can produce fully self-contained JSON string for node graphs.
Such a node graph does not contain any UUID reference (self-contained) and can be super helpful for users if they want to store their work in a JSON file  before uploading it to CRIPT.

@bearmit expressed that he would like this feature.

## Changes

Also added a hint about using `kwargs` to beautify JSON strings into the documentation.

## Tests

A test is included to ensures that self-containing JSON can be parsed by `cript.load_nodes_from_json`.

## Checklist

### DIFF
--- a/src/cript/nodes/core.py
+++ b/src/cript/nodes/core.py
@@ -242,15 +242,161 @@ class BaseNode(ABC):
         self.validate()
         return self.get_json().json
 
-    def get_self_contained_json(self, **kwargs):
+    def get_expanded_json(self, **kwargs) -> str:
         """
-        This generates a JSON representation of the current node at root that is self-contained.
-        Self-contained means in this context that we do not have references via UUID to external node (i.e. present only the CRIPT back end).
-        Such a self-contained JSON can be helpful to temporarily store CRIPT nodes in a file to be uploaded to CRIPT at a later time.
-        A self-contained JSON is also guaranteed to be loadable via `cript.load_nodes_from_json()` into the SDK.
+        Generates a long-form JSON representation of the current node and its hierarchy.
 
-        A self-contained JSON may not be directly compatible with the JSON schema requirements for POST or PATCH uploads.
-        Similar to `get_json` we also accept kwargs, that are passed on to the JSON decoding via `json.dumps()`  this can be used for example to prettify the output.
+        The long-form JSON includes complete details of the node, eliminating the need for
+         references to UUIDs to nodes stored in the CRIPT database. This comprehensive representation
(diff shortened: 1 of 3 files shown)


tests/test_node_util.py
```diff
@@ -323,3 +323,25 @@ def test_invalid_project_graphs(simple_project_node, simple_material_node, simpl
...
+
+def test_expanded_json(complex_project_node):
+    """
+    Tests the generation and deserialization of expanded JSON for a complex project node.
+
+    This test verifies 2 key aspects:
+        1. A complex project node can be serialized into an expanded JSON string, without UUID placeholders.
+        2. The expanded JSON can be deserialized into a node  that is equivalent to the original node.
+    """
+    expanded_project_json: str = complex_project_node.get_expanded_json()
```
nh916, line 336: in hindsight, I probably should have named this `project_expanded_json` to make more sense, but I think it'll pass for now

nh916 (approved):

tests/test_node_util.py
```diff
@@ -323,3 +323,25 @@ def test_invalid_project_graphs(simple_project_node, simple_material_node, simpl
...
+    deserialized_project_node: cript.Project = cript.load_nodes_from_json(expanded_project_json)
+
+    # assert the expanded JSON was correctly deserialized to project node
+    assert deserialized_project_node == complex_project_node
+
+    short_json: str = complex_project_node.json
```
nh916, line 342: this should have also been `condensed_json` instead to make it more self documenting and clearer

src/cript/nodes/core.py
```diff
@@ -242,6 +242,164 @@ def json(self):
...
+        >>> my_inventory = cript.Inventory(
+        ...     name="my inventory", material=[my_material_1, my_material_2]
+        ... )
+        >>> #  ============= Assemble nodes =============
+        >>> my_project.collection = [my_collection]
+        >>> my_project.collection[0].inventory = [my_inventory]
+        >>> #  ============= Get long form JSON =============
+        >>> long_form_json = my_project.get_expanded_json(indent=4)
+
+        ???+ info "Short JSON VS Long JSON"
```
nh916, line 304: We should have called this section `Condensed JSON VS Expanded JSON`, but it is okay for now and we can come back to it later

nh916 (approved):

tests/test_node_util.py
InnocentBug: agreed

src/cript/nodes/core.py
InnocentBug: agreed

PR #422, opened 2023-12-07
DO NOT MERGE disable UID generation

InnocentBug:
# Description

Give an option to disbale the UID use for JSON.

@bearmit 

## Known Issues

Disabling the use of UIDs, make JSON generation for graphs that contain cycles crash.
This is not gracefully handled at the moment.

I am not sure it is a good idea to offer users on option that leads to ungracefully handled errors.

It would be nice to offer this option for @bearmit work.
But I am scared to include it, what do you think @Ardi028 ?

## Checklist

- [y] My name is on the list of contributors (`CONTRIBUTORS.md`) in the pull request source branch.
- [half] I have updated the documentation to reflect my changes.
- [n] My code changes have been verified by automated tests and pass all relevant test scenarios.

### DIFF
--- a/src/cript/nodes/util/__init__.py
+++ b/src/cript/nodes/util/__init__.py
@@ -103,8 +103,9 @@ class NodeEncoder(json.JSONEncoder):
             except AttributeError:
                 pass
             else:
-                if uid in NodeEncoder.handled_ids:
-                    return {"uid": uid}
+                pass
+                # if uid in NodeEncoder.handled_ids:
+                #     return {"uid": uid}
 
             # When saving graphs, some nodes can be pre-saved.
             # If that happens, we want to represent them as a UUID edge only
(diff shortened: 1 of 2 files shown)


bearmit:
Ok, I might do this manually. I suppose I would need to disable the same way the use of EdgeUUID (line 117) too.

bearmit:
@InnocentBug It worked like a charm.

src/cript/nodes/util/__init__.py
```diff
@@ -38,6 +38,8 @@ class NodeEncoder(json.JSONEncoder):
         A set to store the node types that should be condensed to UUID edges in the JSON.
     suppress_attributes : Optional[Dict[str, Set[str]]]
         A dictionary that allows suppressing specific attributes for nodes with the corresponding UUIDs.
+    suppress_uid_usage : bool
+        Can suppress the usage of UID abbreviations. Warning, this can lead to infinite recursion with graphs that contain cycles.
```
bearmit, line 42: Idea: we could have two depth limits and a `depth` argument passed in recursive calls.

For example:
- `fullNodeDepth` would serialize to full nodes until this depth level is reached

Example:

F: full node
E: edge
fullNodeDepth: 4

F*: are the same node, serialized differently due to the backreference

```
F -> F -> F -> F* -> E
   \
     -> F -> F* ->F -> E
           \
              -> F -> F -> E
```

PR #425, opened 2024-02-21
Refactor paginator

InnocentBug:
# Description

This refactors the paginator.
Instead of having the user manually page through the pages of search results, this improved paginator is a python iterator.

So users, can just loop through the results with `for node in paginator` as you would expect from python.
A new change is also that the paginator now directly returns nodes instead of JSON that can be converted to nodes.

## Changes
Tests are adjusted to the new design.

## Known Issues
- 2 search tests fail at the moment.
- One because an invalid material is returned upon search
- One because the bigsmile search connection is timing out.

## Notes
Tagging @duboyal for visibility.

## Checklist


brili (dismissed):

InnocentBug:
Merge #429 first.

src/cript/api/api.py
```diff
@@ -235,7 +232,7 @@ def __str__(self) -> str:
         ...     storage_token=os.getenv("CRIPT_STORAGE_TOKEN")
         ... ) as api:
         ...     print(api)
-        CRIPT API Client - Host URL: 'https://api.criptapp.org/api/v1'
+        CRIPT API Client - Host URL: 'https://api.criptapp.org'
```
duboyal, line 235: i'm curious is this change cascading to other aspects of the code (implications for me) or no ? I will just leave comments here and there to try to thouroughly understand as much as possible

duboyal (approved):
approval for capsule

src/cript/api/paginator.py
```diff
@@ -1,58 +1,46 @@
...
     """
     Paginator is used to flip through different pages of data that the API returns when searching.
     > Instead of the user manipulating the URL and parameters, this object handles all of that for them.
 
-    When conducting any kind of search the API returns pages of data and each page contains 10 results.
-    This is equivalent to conducting a Google search when Google returns a limited number of links on the first page
-    and all other results are on the next pages.
+    Using the Paginator object, the user can simply and easily flip through the results of the search.
+    The details, that results are listed as pages are hidden from the user.
+    The pages are automatically requested from the API as needed.
```
duboyal, line 19: ok so this comment is the heart of the paginator functionality

src/cript/api/data_schema.py
```diff
@@ -20,14 +18,28 @@ class DataSchema:
...
+        --------
+        ### Create a stand alone DataSchema instance.
+        >>> import cript
+        >>> with cript.API(host="https://api.criptapp.org/") as api:
+        ...    data_schema = cript.api.DataSchema(api)
+        """
+        self._api = api
+        self._vocabulary = {}
+        self._db_schema = self._get_db_schema()
+
```
duboyal, line 41: is this a constructor for a data schema class where it preloads all the schemas for the nodes and make that accessible throughout other parts of the code? did this constructor not really exist before ?

InnocentBug: it is not really a constructor, since it is just getting the JSON data as a dict from the API.
This call existed before, but was in a different location.
I moved it now into the constructor of the DataSchema class, since a DataSchema without loaded JSON schema is not useful.

src/cript/api/paginator.py
```diff
@@ -1,58 +1,46 @@
...
         self,
-        http_headers: dict,
-        api_endpoint: str,
-        query: Optional[str] = None,
-        current_page_number: int = 0,
+        api,
+        url_path: str,
+        page_number: Union[int, None],
+        query: str,
     ):
```
duboyal, line 44: does this object init function take a page number ? just curious in a use case of this instantiation how does a user know what page number to pass in unless there is also a way to get a number of pages?

InnocentBug: It does take a page number, because some queries do not allow pages to be specified. 
So in that case a user can specify None for the page. 

Also the constructor of the class is mostly hidden from the user as only api.search instantiates them and returns it to the user. 

But just in case, there is default so it's transparent for users.

duboyal: @InnocentBug  could you give a quick example where the call to the paginator would look like to retrieve an already existing node? I left this comment on my sdk update as well because you left an original comment to be implementing api.search or paginator .

PR #428, opened 2024-02-23
Sdk update

duboyal:
# Description
this allows nodes to be changed and saved based off the deep diff registry of changes 

I will be updating the description on a more technical basis before merging
## Checklist

- [ ] My code changes have been verified by automated tests and pass all relevant test scenarios.


script_cript.py
InnocentBug, line 1: This file should not be part of the of the PR

src/cript/api/api.py
```diff
@@ -31,7 +34,12 @@
 )
 from cript.api.utils.web_file_downloader import download_file_from_url
 from cript.api.valid_search_modes import SearchModes
+from cript.nodes.primary_nodes.primary_base_node import PrimaryBaseNode
+
 from cript.nodes.primary_nodes.project import Project
+from deepdiff import DeepDiff
+import cript
```
InnocentBug, line 41: you can't import cript here.
That's a circular import.

Be specific and import exactly what you need.

src/cript/api/api.py
```diff
@@ -48,6 +56,16 @@ def _get_global_cached_api():
     return _global_cached_api
 
 
+# load from .env
```
InnocentBug, line 59: not a good idea, this should be the API class.

src/cript/api/api.py
```diff
@@ -484,6 +502,404 @@ def _check_initial_host_connection(self) -> None:
         except Exception as exc:
             raise CRIPTConnectionError(self.host, self._api_token) from exc
 
+    # ================ helpers ==============
+    @classmethod
+    def object_exists(cls, node: str, name: str):
```
InnocentBug, line 507: use the api.search function to see if an object exist

src/cript/api/api.py
```diff
@@ -484,6 +502,404 @@ def _check_initial_host_connection(self) -> None:
...
+                return True, item.get("uuid")
+
+            return False, None
+
+        except requests.RequestException as e:
+            print(f"An error occurred: {e}")
+            return False, None
+
+    @staticmethod
+    def fetch_object_data(object_type: str, uuid: str):
```
InnocentBug, line 535: should be through paginator and search.

src/cript/api/api.py
```diff
@@ -484,6 +502,404 @@ def _check_initial_host_connection(self) -> None:
...
+        response.raise_for_status()
+        if response.json()["data"]:
+
+            return response.json()["data"][0]  # Assuming the first item is the desired one
+
+        else:
+            raise ValueError(f"No data available in response {response}")
+
+    @staticmethod
+    def load_schema_for_node_type(node_type):
```
InnocentBug, line 554: not necessary.
Node testing is already implemented.

src/cript/api/api.py
```diff
@@ -484,6 +502,404 @@ def _check_initial_host_connection(self) -> None:
...
+        for item in node_objects:
+            if item["name"] in duplicate_list:
+                removed_items.append(item)
+            else:
+                remaining_items.append(item)
+
+        return remaining_items
+
+    @staticmethod
+    def send_patch_request_for_node(parent_node=None, payload=None):  # (self, base_url, entity_uuid, data, auth_token=None):
```
InnocentBug, line 607: this should not be a public function.
Users should never call this function.

Also, shouldn't be a static function, host, header etc. should be loaded from api class.

src/cript/api/api.py
```diff
@@ -484,6 +502,404 @@ def _check_initial_host_connection(self) -> None:
...
+        data = payload
+        response = requests.patch(url, json=data, headers=headers)
+        if response.status_code == 200:
+            print("\nEntity updated successfully.", response.status_code)
+        else:
+            print(f"Entity patch updated Failed. Status code: {response.status_code}, Response: {response.text}")
+        return response
+
+    @staticmethod
+    def send_delete_request_for_node(parent_node=None, payload=None):  # (self, base_url, entity_uuid, data, auth_token=None):
```
InnocentBug, line 628: should be `delete` function of API.

Also not a static function.

src/cript/api/api.py
```diff
@@ -484,6 +502,404 @@ def _check_initial_host_connection(self) -> None:
...
+        data = payload
+        response = requests.delete(url, json=data, headers=headers)
+        if response.status_code == 200:
+            print("\nEntity updated successfully.", response.status_code)
+        else:
+            print(f"Entity delete updated Failed. Status code: {response.status_code}, Response: {response.text}")
+        return response
+
+    @staticmethod
+    def find_uuid_by_name_and_type(name, node_type):
```
InnocentBug, line 649: use search and paginator.

src/cript/api/api.py
```diff
@@ -484,6 +502,404 @@ def _check_initial_host_connection(self) -> None:
...
+                print(search_url)
+                return None
+        except requests.RequestException as e:
+            print(f"Error querying UUID for name '{name}' and type '{node_type}': {e}")
+            print("search_url")
+            print(search_url)
+            return None
+
+    @classmethod
+    def create_node(cls, node_type, data):
```
InnocentBug, line 706: should not be public.
Maybe and should be capsuled in `api.save`

src/cript/api/api.py
```diff
@@ -484,6 +502,404 @@ def _check_initial_host_connection(self) -> None:
...
+        if node_type == "project":  # elif node type is something else it will have a parent
+            headers = {"Authorization": f"Bearer {Config.api_token}", "Content-Type": "application/json"}
+            url = "https://lb-stage.mycriptapp.org/api/v1/project/"
+
+            response = requests.post(url, json=data, headers=headers)
+            return response
+
+    @classmethod
```
InnocentBug, line 714: not a pulic API class.
Maybe in api.utils as private function if needed.

InnocentBug (changes requested):

src/cript/api/api.py
InnocentBug: I realize, that this comment is not helpful at the moment, because I am refactoring the paginator.
but eventually, we should strive to not duplicate code paths that essentially do the same thing....

src/cript/api/api.py
duboyal: I think here i will implement the capsule to resolve this

src/cript/api/api.py
duboyal: same here - implement capsule

src/cript/api/api.py
duboyal: @InnocentBug  could you give a quick example where the call to the paginator would look like to retrieve an already existing node?

InnocentBug: Sure.
```python
import cript

with cript.API() as cript_api: # Assumes you have the host and token in the environment.
    # If you know its uuid
    uuid_paginator = cript_api.search(node_type=cript.Project, search_mode=cript.SearchModes.UUID, value_to_search="e7381525-64b9-4849-a416-57aa1e09044f")
    project = next(uuid_paginator)
    print(project)

    # if you know its name
    exact_name_paginator = cript_api.search(node_type=cript.Material, search_mode=cript.SearchModes.EXACT_NAME, value_to_search="Sodium polystyrene sulfonate")
    material = next(exact_name_paginator)
    print(material)
```

PR #429, opened 2024-02-23
Capsule requests

InnocentBug:
# Description

The idea is to never call `requests.XXX` directly, but to use `cript.API._capsule_request` instead.
This has the advantage that we only make direct calls to the CRIPT API from a single point. And only through the API object.
This ensure we only hit the right host.

But the biggest advantage is, that we can consistently log all calls and responses to the API.
This should help debugging, and make communicating with the API team much easier, since we always have the exact API call, and the response we are working with logged.
No more extra digging to find the repsonse.

## Changes

api.host only returns the basename URL now.
api.api_prefix = "api/" and `api.api_version=v1` are not automatically attached anymore. 

## Known Issues

This is implemented for the paginator or the file uploader yet.
(So those API calls go unlogged.)
I will merge this into #425 and address the paginator there.

The file upload needs a refactor of the design too, and we will address logging there.

## Notes

Happy to receive feedback for this idea.
@duboyal it would be great if we can have the new save work with this too at some point.

## Checklist

### DIFF
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -232,7 +232,7 @@ class API:
         ...     storage_token=os.getenv("CRIPT_STORAGE_TOKEN")
         ... ) as api:
         ...     print(api)
-        CRIPT API Client - Host URL: 'https://api.criptapp.org/'
+        CRIPT API Client - Host URL: 'https://api.criptapp.org'
 
         Returns
         -------
@@ -367,7 +367,7 @@ class API:
         ...     storage_token=os.getenv("CRIPT_STORAGE_TOKEN")
         ... ) as api:
         ...    print(api.host)
-        https://api.criptapp.org/
+        https://api.criptapp.org
         """
         return self._host
 
--- a/src/cript/api/api.py
+++ b/src/cript/api/api.py
@@ -2,6 +2,7 @@ import copy
 import json
 import logging
 import os
+import traceback
 import uuid
 from pathlib import Path
 from typing import Any, Dict, Optional, Union
@@ -71,6 +72,8 @@ class API:
     _BUCKET_DIRECTORY_NAME: str = "python_sdk_files"
     _internal_s3_client: Any = None  # type: ignore
     # trunk-ignore-end(cspell)
+
+    extra_api_log_debug_info: bool = False
 
     @beartype
     def __init__(self, host: Union[str, None] = None, api_token: Union[str, None] = None, storage_token: Union[str, None] = None, config_file_path: Union[str, Path] = "", default_log_level=logging.INFO):
@@ -211,9 +214,6 @@ class API:
 
         # add Bearer to token for HTTP requests
         self._http_headers = {"Authorization": f"Bearer {self._api_token}", "Content-Type": "application/json"}
-
-        # check that api can connect to CRIPT with host and token
-        self._check_initial_host_connection()
 
         # set a logger instance to use for the class logs
         self._init_logger(default_log_level)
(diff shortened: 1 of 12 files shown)


src/cript/api/api.py
```diff
@@ -235,7 +232,7 @@ def __str__(self) -> str:
         ...     storage_token=os.getenv("CRIPT_STORAGE_TOKEN")
         ... ) as api:
         ...     print(api)
-        CRIPT API Client - Host URL: 'https://api.criptapp.org/api/v1'
+        CRIPT API Client - Host URL: 'https://api.criptapp.org'
```
duboyal, line 235: moving my comment from the other PR:

i'm curious is this change cascading to other aspects of the code (implications for me) or no ? I will just leave comments here and there to try to thouroughly understand as much as possible

src/cript/api/api.py
```diff
@@ -497,7 +444,7 @@ def _internal_save(self, node, save_values: Optional[_InternalSaveValues] = None
 
             # This checks if the current node exists on the back end.
             # if it does exist we use `patch` if it doesn't `post`.
-            test_get_response: Dict = requests.get(url=f"{self._host}/{node.node_type_snake_case}/{str(node.uuid)}/", headers=self._http_headers, timeout=_API_TIMEOUT).json()
+            test_get_response: Dict = self._capsule_request(url_path=f"/{node.node_type_snake_case}/{str(node.uuid)}/", method="GET").json()
```
duboyal, line 447: ok this is like the heart of the PR

InnocentBug: yes!

src/cript/api/api.py
```diff
@@ -997,11 +948,57 @@ def delete_node_by_uuid(self, node_type: str, node_uuid: str) -> None:
...
-        response: Dict = requests.delete(headers=self._http_headers, url=delete_node_api_url, timeout=_API_TIMEOUT).json()
+        response: Dict = self._capsule_request(url_path=f"/{node_type.lower()}/{node_uuid}/", method="DELETE").json()
 
         if response["code"] != 200:
-            raise APIError(api_error=str(response), http_method="DELETE", api_url=delete_node_api_url)
+            raise APIError(api_error=str(response), http_method="DELETE", api_url=f"/{node_type.lower()}/{node_uuid}/")
 
         self.logger.info(f"Deleted '{node_type.title()}' with UUID of '{node_uuid}' from CRIPT API.")
+
+    def _capsule_request(self, url_path: str, method: str, api_request: bool = True, headers: Optional[Dict] = None, timeout: int = _API_TIMEOUT, **kwargs) -> requests.Response:
```
duboyal, line 959: ok - this is the heart of the PR

duboyal: basically would you say with this change brought in I should use capsule each time instead of any manual requests.post/patch/delete

src/cript/api/api.py
InnocentBug: It should not have no effect (or a manageable effect) on the cascade downstream.

Since most API requests are capsuled after this, the capsuled request autmotically appends this for you.
(It can be optionally disabled.)
And that should handle %95 of all situations.

If it is not handled by this, you can recover the old behavior by

`hostname = f"{api.host}/{api.api_prefix}/{api.api_version}"`

src/cript/api/api.py
InnocentBug: yes, that is what it means.

It just makes debugging so much easier, because we can log the requests and send them to the back end team directly, without manually finding and reproducing the error.

src/cript/api/data_schema.py
```diff
@@ -63,23 +73,7 @@ def _get_db_schema(self, host: str) -> dict:
...
-        ### Create a stand alone DataSchema instance.
-        >>> import cript
-        >>> with cript.API(host="https://api.criptapp.org/") as api:
-        ...    data_schema = cript.api.DataSchema(api.host)
-        """
-
-        self._db_schema = self._get_db_schema(host)
-        self._vocabulary = self._get_vocab(host)
-        self._logger = logger
-
```
duboyal, line 81: 👍

src/cript/nodes/uuid_base.py
```diff
@@ -52,7 +52,7 @@ def url(self):
         from cript.api.api import _get_global_cached_api
 
         api = _get_global_cached_api()
-        return f"{api.host}/{self.uuid}"
+        return f"{api.host}/{api.api_prefix}/{api.api_version}/{self.uuid}"
```
duboyal, line 55: ahh ok this compensates for it being removed in the other part of the codebase

InnocentBug: yes!

InnocentBug:
Has been merged with #425

PR #440, opened 2024-03-11
Add a working patch option for `get_json`

InnocentBug:
# Description

Minimal change, but removes the `uuid` field from the top-level node in a patch call.

## Checklist

### DIFF
--- a/src/cript/nodes/core.py
+++ b/src/cript/nodes/core.py
@@ -439,6 +439,7 @@ class BaseNode(ABC):
         @dataclass(frozen=True)
         class ReturnTuple:
             json: str
+            json_dict: dict
             handled_ids: set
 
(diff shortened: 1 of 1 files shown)


InnocentBug:
Ok, this is the second part of the puzzle that I promised you.

In order to create JSON for the reverse DFS patches, here is what you can do:

```python
# create a set of all, old known uuids, but exclude the node to patch
suppress_uuid = set(old_tree_map.keys()) - {str(node.uuid)}
patch_json = node.get_json(is_patch = True, known_uuid=suppress_uuid)
```
I assume you have the `old_tree_map` as discussed.

The `is_patch` now removes the `uuid` from the top level node, I think that is required at the moment.

duboyal (approved):

PR #441, opened 2024-03-12
Fix problems with the node cache

InnocentBug:
# Description

Now the node UUID cache actually works as intended.
Users can only have one python object with the same UUID.

## Changes

Added a functionality that allows you to circumvent this cache explictly.
This should be a feature that users don't really use.

## Checklist

### DIFF
--- a/src/cript/nodes/util/json.py
+++ b/src/cript/nodes/util/json.py
@@ -14,6 +14,7 @@ from cript.nodes.exceptions import (
     CRIPTJsonDeserializationError,
     CRIPTJsonNodeError,
 )
+from cript.nodes.uuid_base import UUIDBaseNode
 
 
 class NodeEncoder(json.JSONEncoder):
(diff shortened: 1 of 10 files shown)


src/cript/nodes/util/json.py
```diff
@@ -349,7 +350,22 @@ def load_nodes_from_json(nodes_json: Union[str, Dict]):
     # but catches a lot of odd cases. And at the moment performance is not bottle necked here
     if not isinstance(nodes_json, str):
         nodes_json = json.dumps(nodes_json)
-    return json.loads(nodes_json, object_hook=node_json_hook)
+
```
InnocentBug, line 353: :heart_decoration:

src/cript/nodes/uuid_base.py
```diff
@@ -32,16 +29,26 @@ class JsonAttributes(BaseNode.JsonAttributes):
 
     _json_attrs: JsonAttributes = JsonAttributes()
 
+    def __new__(cls, *args, **kwargs):
+        uuid: Optional[str] = kwargs.get("uuid")
```
InnocentBug, line 33: this fixes the bug that we had before.

Fun playing with something new like the `__new__`

duboyal (approved):

duboyal (approved):

PR #444, opened 2024-03-20
init the solution to get cognito tokens

izaim:
## Changes
added ability to get tokens from cognito

## Known Issues
no issues


InnocentBug:
This looks fine to me on first glance.
Any reason it is closed?

PR #449, opened 2024-04-01
Update newest (no deep diff)

duboyal:
# Description

## Checklist

- [ ] My code changes have been verified by automated tests and pass all relevant test scenarios.


tests/nodes/primary_nodes/test_project.py
```diff
@@ -95,3 +98,46 @@ def test_integration_project(cript_api, simple_project_node):
...
+    proj0 = copy.deepcopy(complex_project_node)
+    proj_json = proj0.get_json().json
+    cript_api.save_new(proj0)
+
+    proj = cript.load_nodes_from_json(nodes_json=json.dumps(proj_json))
+    print("\n----proj_loaded")
+    print(proj)
+
+    print("------------\n1111111")
+    print(type(proj))
```
duboyal, line 125: @InnocentBug Ludwig, if you have a moment do you mind telling me why this might load in a string?

also my work flow and logic in this test , is to ensure that I can save the node as just a "creation" step , and then be able to reload the node back in ( via using "load nodes from json") , 

then the idea to edit it , save it , then load it again ensuring the changes are indeed saved,
but the return value should be a node object right ? this is also what the function docs say and what I remember in my past work. I made this branch off of the current development and that function has not been touched by the changes in this PR so I'm a bit perplexed though its probably something simple , so I'm wondering what your thoughts are

InnocentBug: It does load a string if it thinks that it isn't a node.
(Or valid JSON)

tests/nodes/primary_nodes/test_project.py
```diff
@@ -95,3 +98,46 @@ def test_integration_project(cript_api, simple_project_node):
...
+    #         proj_json = json.load(json_handle)
+
+    # proj = cript_api.load_nodes_from_json(nodes_json=proj_json)
+    # Modify deep in the tree
+    print("------------\nstarting")
+    proj0 = copy.deepcopy(complex_project_node)
+    proj_json = proj0.get_json().json
+    cript_api.save_new(proj0)
+
+    proj = cript.load_nodes_from_json(nodes_json=json.dumps(proj_json))
```
InnocentBug, line 120: also be careful here, if you load an existing project with the same UUID.
It will give you the same nodes as you have before.

You can't have different python nodes with the same UUID, except if you use the `_use_cache` argument of the load function.

InnocentBug:
I just realized, that we may get a number of false positive `equal_shallow` and `equal_deep` calls.
The API adds info to nodes, like `created_at` etc. and these don't actually mean a diff, but will trigger a patch in the current design.

I don't think that is an issue, but something to keep in mind, and maybe for optimization later.

src/cript/api/api.py
```diff
@@ -402,7 +424,114 @@ def api_prefix(self):
...
+            # Do the POST request call. only on project
+            # or else its a patch handled by previous node
+
+            print("--------900---------")
+            print(new_node.node_type.lower())
+            if new_node.node_type.lower() == "project":
+                data = new_node.get_json().json
+                response = self._capsule_request(url_path="/project/", method="POST", data=data)
+                print("----700----")
+                print(response.json())
```
duboyal, line 466: @InnocentBug 

Hey Ludwig 
this is giving me the "bad uuid" for a project when I run "test inventory"

I get this out put 


```
--------900---------
project
----700----
{'code': 400, 'data': None, 'error': 'Bad uuid: 9aa4b9e7-1b7c-4c76-9602-2c980bc4298a provided'}
```

Screenshot 2024-04-02 at 5 43 51 PM

InnocentBug: That is probably the same issue that we discussed before.

The API requires all materials to be known before an inventory can be saved.
But in our tests, we have inventories with new materials in them.

So, you get a bad UUID, when you try to save one.

We will have to redesign the fixtures, such that they contain known materials.

InnocentBug: for now, we can ignore it, I guess.

Or better, change the inventories, such that they contain no or known materials.

But we also have to think about the user, parse this error, and give a good error message, when this happens.

src/cript/api/api.py
```diff
@@ -431,14 +431,26 @@ def save(self, new_node):
         quit()
         self._internal_save(new_node, _no_condense_uuid=False)
 
-    def _internal_save(self, new_node: PrimaryBaseNode, _no_condense_uuid: bool) -> None:
-        print("_no_condense_uuid")
-        print(_no_condense_uuid)
+    def _internal_save(self, new_node: PrimaryBaseNode, preknown_uid: str, _no_condense_uuid: bool) -> None:
+        """
+        NOTE: for Ludwig
```
duboyal, line 436: @InnocentBug 

Ludwig , this is still Work In Progress, I think I fleshed out most of the skeleton but where to put and use the "uid map" is causing me to scratch my head a bit. I left a pretty detailed note in the docstring here. do you mind taking a a glance at it ? (the note?) and let me know your thoughts?

PR #454, opened 2024-04-24
Update to 1.0.1

InnocentBug:


duboyal (approved):

tests/fixtures/primary_nodes.py
```diff
@@ -25,7 +25,7 @@ def simple_project_node(simple_collection_node) -> cript.Project:
 def complex_project_dict(complex_collection_node, simple_material_node, complex_user_node) -> dict:
     project_dict = {"node": ["Project"]}
     project_dict["locked"] = True
-    project_dict["model_version"] = "1.0.0"
+    project_dict["model_version"] = "1.0.1"
```
brili, line 28: I would advise you dont pin the `model_version` to a specific version as it is populated automatically on the latest version.

InnocentBug: great idea, how do we know which one is the newest version?

InnocentBug: also this is probably only a work around.

We need a general scheme to update nodes from one version to the newest version automatically.

In this case this is very simple, but it can get more complicated easily.

Do you have ideas how this can be done?

I was thinking something along the Go philosophy that any change needs to have an automatic port option.
Otherwise we will have nodes that we can't process anymore since they are going out of date.

brili: You can get the latest version from the `/schema` endpoint at `#/defs/ModelVersion", but I think you are looking too much into it, I think the datamodel as of now wont have major changes that will need transformations, the only changes that will be periodical is from the controlled vocabulary which is also embedded in the jsonschema hence bumping the version number. It should work fine if you remove the model_version attribute altogether.

InnocentBug: can you approve pls

PR #456, opened 2024-05-14
Update search, so it works with the new `after` syntax.

InnocentBug:
# Description
Updated search URL generation.
It now works with 

## Changes

Some updates to the paginator interface that enables partial searches, that allows interrupted searches for the SDK as well.

## Known Issues

Many entries on `stage` are invalid.
So we had to work around this in the tests.
Ideally we don't have to do that.

@brili 
I noticed, that BigSMILES search does not deliver 10 nodes by default. It is 1 node for me.

@brili 
Also it appears that the `after` method returns some nodes multiple times.
I get the same UUID in the search results, I don't think that is as it should be.

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,7 +35,7 @@ jobs:
 
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.8, 3.12]
+        python-version: ["3.10", 3.12]
 
     env:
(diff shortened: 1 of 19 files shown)


brili (approved):

brili:
@InnocentBug thanks for pointing out the dupes issue, its an easy fix.

brili (approved):

